PR #51, opened 2026-03-23
Add subprocess provider plugins

hughhan1:
This adds a `plugin` integration backend that lets Gestalt launch a managed child process and treat it like a normal provider, so custom integrations written in TypeScript, Python, or Rust are invokable through the existing REST and MCP surfaces.

This also documents the `gestalt-plugin/1` protocol, includes starter SDKs for TypeScript, Python, and Rust, and adds end to end coverage for execution, manual auth, and OAuth so the cross language path is exercised with the same provider semantics as in process Go implementations.


hughhan1:
Closing in favor of the gRPC pluginapi from PRs #95-96, which covers both providers and runtimes with host callbacks. The OAuth error-propagation fixes identified during review will be tracked separately.

PR #61, opened 2026-03-23
Warn on risky SQLite deployment configs

hughhan1:
## Summary
- add SQLite datastore diagnostics for local paths, temp paths, and deployment environments that should stay single-instance
- print those warnings during normal startup and in `gestaltd --check`
- document the warning behavior and clarify the expected SQLite deployment model in config and deploy docs

## Testing
- go test ./internal/config/...
- go test ./cmd/gestaltd/...
- go test ./...


hughhan1:
Closing this because it was opened against  by mistake. Replacement PR: #62, opened against  with only the SQLite warning change.

PR #95, opened 2026-03-24
Add pluginapi gRPC contract and MCP OAuth tests

hughhan1:
This introduces the plugin API, a gRPC-based contract that lets
provider and runtime plugins run as separate executables while
communicating with Gestalt over a well-defined protocol. The protobuf
schema covers provider operations (invoke, list capabilities) and
runtime lifecycle (start, stop, host callbacks for invoking other
providers and listing capabilities). Go adapter layers wrap the
generated gRPC stubs so plugins can implement plain Go interfaces
without touching protobuf directly.

Separately, this adds functional tests for the MCP OAuth discovery and
Dynamic Client Registration flow, covering both the RFC 9728 indirection
path and the direct-endpoints path against a mock server.


hughhan1:

- #95 ◀
    - #96

PR #96, opened 2026-03-24
Load provider and runtime plugins from executables

hughhan1:
Integrations and runtimes can now be backed by external executables
instead of compiled-in factories. Setting a `plugin:` block in the
config with a `command` (and optional `args`/`env`) causes Gestalt to
spawn the executable as a child process and communicate with it over
the pluginapi gRPC protocol. This lets plugin authors write providers
and runtimes in any language without linking against Gestalt internals.

The bootstrap layer manages the full lifecycle: starting the process,
waiting for the gRPC socket, and tearing down gracefully on shutdown.
An echo plugin binary exercises the end-to-end path in tests, including
a runtime that calls back into the host to invoke a provider.

### DIFF
--- a/cmd/gestalt-plugin-echo/main.go
+++ b/cmd/gestalt-plugin-echo/main.go
@@ -1,0 +1,133 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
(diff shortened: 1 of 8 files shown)


hughhan1:

- #95
    - #96 ◀

PR #166, opened 2026-03-26
Add code-host providers for GitHub, GitHub App, and GitLab

hughhan1:
These three new first-party providers give Gestalt native support for
the most common code hosting platforms. The `github` provider uses
OAuth2 for user-level access, `github_app` uses installation auth for
app-level access, and `gitlab` uses OAuth2 with PKCE. Each provider
defines a comprehensive set of API operations covering repositories,
issues, pull/merge requests, CI/CD, and code search.

The `github_app` provider uses installation-specific endpoints
(`/installation/repositories`, `/app/installations/{id}`) rather than
user-scoped endpoints that are inaccessible to app installation tokens.
The `gitlab` provider splits file management into separate `create_file`
(POST) and `update_file` (PUT) operations to match GitLab's API
semantics. This also removes stale `configure-jira` and
`configure-bigquery` entries from the docs navigation that were breaking
the docs build.

### DIFF
--- a/cmd/gestaltd/providers.go
+++ b/cmd/gestaltd/providers.go
@@ -3,10 +3,16 @@ package main
 import (
 	"github.com/valon-technologies/gestalt/internal/bootstrap"
 	"github.com/valon-technologies/gestalt/plugins/providers/bigquery"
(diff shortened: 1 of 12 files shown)


hughhan1:

PR #168, opened 2026-03-26
Add first-party Google REST providers

hughhan1:
This adds four new declarative REST providers for Google Workspace
services: `google_calendar`, `google_docs`, `google_drive`, and
`google_sheets`. Each provider uses OAuth2 with offline access for
token refresh and covers the core operations for its respective
Google API, giving agents built-in access to calendars, documents,
files, and spreadsheets without requiring external plugin installation.


hughhan1:

PR #169, opened 2026-03-26
Add ops and incident provider definitions

hughhan1:
This adds first-party providers for Datadog, incident.io, LaunchDarkly,
and PagerDuty covering monitoring, incident management, feature flags,
and on-call operations. Each provider follows the existing YAML
definition pattern with factory, tests, and registration in the central
provider registry.


hughhan1:

PR #170, opened 2026-03-26
Add dbt Cloud, Hex, Intercom, and Notion providers

hughhan1:
This adds four new workflow and knowledge providers, each with a
provider YAML definition, a factory, and factory tests. The Intercom
search operation uses the nested query object structure required by the
Intercom API. Stale doc nav entries for removed Jira and BigQuery
configuration pages are also cleaned up to fix the docs build.


hughhan1:

PR #172, opened 2026-03-26
Add first-party providers for Extend, Figma, Gong, Ramp, and Rippling

hughhan1:
These five new providers expand the set of built-in integrations
available without user-supplied OpenAPI specs or upstream configuration.
Extend covers document processing, Figma covers design collaboration,
Gong covers conversation intelligence, Ramp covers corporate expense
management, and Rippling covers HR and employee management. Each
provider ships a YAML definition with full operation coverage and is
registered in the main factory so it works out of the box.


hughhan1:
Closing: definition-only providers will be re-implemented as OpenAPI spec configs in valon-tools instead of hand-authored provider.yaml files. See AIT-232.

PR #239, opened 2026-03-27
Add builder target to Dockerfile

hughhan1:
The existing Dockerfile only produces a minimal distroless image with the gestaltd binary, which is not usable as a base for downstream plugin builds. This adds a builder target that retains the Go toolchain, git, and module cache so that downstream Dockerfiles can run go build, gestaltd plugin package, and gestaltd bundle in a single stage without pulling in a separate Go image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,8 @@ RUN npm ci
 COPY web/ .
 RUN npm run build
 
-FROM golang:1.26-alpine AS builder
+FROM golang:1.26-alpine AS build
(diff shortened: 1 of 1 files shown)


hughhan1:
Closing this. The builder target doesn't add enough value over the pattern already shown in #238: `FROM golang:1.26-alpine` + `COPY --from=valontechnologies/gestalt:latest /gestaltd /usr/local/bin/gestaltd`. That approach also lets downstream users choose their own Go version.

PR #347, opened 2026-03-30
Fix CLI credential file races and OAuth login timeout

hughhan1:
Credential file writes now set 0o600 permissions atomically on Unix instead of writing first and then changing permissions, closing a window where the file was briefly world-readable. The `load()` and `delete()` methods no longer check existence before acting, eliminating TOCTOU race conditions where the file could appear or disappear between the check and the operation.

The OAuth login flow now times out after 5 minutes instead of blocking indefinitely, so the CLI no longer hangs when the user abandons the browser authentication.


hughhan1:
Superseded by #349 which includes the permissions fix for pre-existing files.

PR #348, opened 2026-03-30
Add singleflight deduplication for OAuth token refresh

hughhan1:
When concurrent requests share an expired token, they all attempt to refresh it simultaneously, risking refresh token rotation conflicts with the OAuth provider. This wraps the refresh call in a singleflight.Group keyed on user/provider/connection/instance so only one goroutine performs the actual refresh. Shared callers re-fetch the token from the datastore to pick up the winner's stored result rather than using the response directly.

The error path continues to re-fetch from the store unconditionally (as before) to handle the case where a separate request already refreshed successfully outside the singleflight window.

### DIFF
--- a/server/go.mod
+++ b/server/go.mod
@@ -44,6 +44,7 @@ require (
 	golang.org/x/crypto v0.48.0
 	golang.org/x/mod v0.34.0
 	golang.org/x/oauth2 v0.36.0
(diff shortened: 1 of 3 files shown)


hughhan1:
Singleflight adds disproportionate complexity for this problem. Replacing with a simpler check-before-write approach.

PR #352, opened 2026-03-30
Improve server error handling and observability

hughhan1:
Several error paths in the server silently discarded errors or logged
them at inappropriate levels. JSON marshal failures during connection
setup and discovery are now surfaced as proper errors instead of being
ignored. The registration store migration failure is logged at error
level rather than warn, and token refresh persistence failures are now
logged with context. Auth middleware logs authentication failures to
aid debugging, and startup warns when no encryption key is configured.


hughhan1:
All meaningful changes from this PR were absorbed during rebase — the remaining diff is just a variable rename. Closing as unnecessary.

PR #354, opened 2026-03-31
Guard against concurrent token refresh overwrites

hughhan1:
When multiple requests share the same expired integration token, each
independently refreshes it with the OAuth provider. If the provider
rotates refresh tokens on each grant, the last writer silently
invalidates the earlier refresh tokens. This adds a re-read from the
store after a successful refresh to detect whether another request
already persisted a new token, avoiding the redundant overwrite.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


hughhan1:
Check-before-write is insufficient — still has a race window. Replacing with singleflight around refreshOAuth.

PR #361, opened 2026-03-31
Add spec-loading and multi-connection support to plugin manifests

hughhan1:
Plugin manifests can now declare `openapi`, `graphql_url`, and `mcp_url` fields instead of hand-writing operations. When a spec URL is set, the server loads and processes the spec at bootstrap time and builds a full provider from it, including composite assembly when both API and MCP URLs are present. An `allowed_operations` map lets manifests filter and override the operations exposed by the loaded spec.

Multi-connection support lets a single plugin declare multiple named auth endpoints via a `connections` map, with `openapi_connection`, `graphql_connection`, and `mcp_connection` bindings to route each surface to its connection. Per-connection OAuth credentials are resolved from plugin config, falling back to top-level `client_id`/`client_secret`. Validation ensures all connections share the same mode and that binding references are valid.

### DIFF
--- a/server/internal/bootstrap/bootstrap.go
+++ b/server/internal/bootstrap/bootstrap.go
@@ -14,8 +14,11 @@ import (
 	"github.com/valon-technologies/gestalt/server/core/integration"
 	"github.com/valon-technologies/gestalt/server/internal/composite"
 	"github.com/valon-technologies/gestalt/server/internal/config"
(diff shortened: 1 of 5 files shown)


hughhan1:
Superseded by #376, which includes all spec-loading and multi-connection changes plus the full plugin-only IntegrationDef simplification.

PR #365, opened 2026-03-31
Simplify IntegrationDef to plugin-only model

hughhan1:
Plugin manifests can now declare an OpenAPI, GraphQL, or MCP URL instead
of hand-writing operations. When a manifest sets one of these fields,
the server fetches the spec at startup and builds a provider from it
automatically, including OAuth connection setup. This lets plugin authors
ship integrations by pointing at a spec rather than coding each
operation individually.

Manifests can also declare multiple named connections with independent
OAuth credentials, so a single plugin can authenticate against several
upstream services. A new `connections` map in the manifest carries per-
connection auth configuration, and the plugin config accepts matching
credential blocks keyed by connection name.


hughhan1:
Closing in favor of #376 which is the more recent version of this change.

PR #383, opened 2026-03-31
Add HTTP proxy host function for plugin processes with AllowedHosts enforcement

hughhan1:
Plugins previously had no way to make outbound HTTP requests through the host,
and the AllowedHosts config field was never enforced. This adds a PluginHost
gRPC service with a ProxyHTTP RPC that all plugin processes (providers and
runtimes) can call back to, using a new GESTALT_PLUGIN_HOST_SOCKET env var.
The host socket is now created for every plugin process, not just runtimes.

When AllowedHosts is configured on a plugin definition, the proxy validates
the target hostname before forwarding the request, supporting both exact
matches and wildcard prefixes (e.g. "*.example.com"). An empty AllowedHosts
list permits all hosts. The SDK gains a ProxiedTransport (http.RoundTripper)
that transparently routes HTTP through the proxy, and the echo test plugin
gains a proxy_http operation used by four new integration tests covering
allowed hosts, blocked hosts, open access, and wildcard matching.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>


hughhan1:
Replacing with Pattern A (transparent HTTP proxy via env vars + OS-level network isolation) instead of Pattern B (host-mediated gRPC API). Pattern A is the industry standard used by Claude Code, Codex, and Cursor — plugins keep using standard HTTP libraries unchanged.

PR #399, opened 2026-03-31
Restrict plugin socket temp directory permissions

hughhan1:
## Summary
- Plugin gRPC sockets in `/tmp` were created with default umask permissions (typically 0022), making the directory world-readable on multi-tenant hosts.
- Adds `os.Chmod(dir, 0700)` after `os.MkdirTemp` to restrict access to the owning user only.

## Test plan
- [x] `go build ./internal/pluginhost/` passes
- [ ] Verify on Linux that `ls -la /tmp | grep gstp` shows `drwx------` after starting a plugin


hughhan1:
Closing — os.MkdirTemp already creates directories with mode 0700, so the chmod is redundant. The original code was secure as-is.

PR #465, opened 2026-04-01
Merge credential field overrides by name

hughhan1:
## Summary
- merge `plugin.auth.credentials` overrides by credential `name` instead of replacing the entire list
- preserve packaged fields when a local override only sets one field such as `description`
- append local-only credential entries that do not exist in the base auth config

## Testing
- `go test ./internal/config`
- `go test ./internal/provider -run 'TestBuildCredentialFields'`
- `go test ./internal/server -run 'TestListIntegrations_ConnectionInfosIncludeConfigCredentialFields|TestListIntegrations_ConnectionInfosIncludeProviderManualAuth'`

### DIFF
--- a/server/internal/config/config.go
+++ b/server/internal/config/config.go
@@ -336,10 +336,55 @@ func MergeConnectionAuth(dst *ConnectionAuthDef, src ConnectionAuthDef) {
 		dst.TokenMetadata = src.TokenMetadata
 	}
 	if len(src.Credentials) > 0 {
-		dst.Credentials = src.Credentials
+		dst.Credentials = mergeCredentialFields(dst.Credentials, src.Credentials)
 	}
 	if src.AuthMapping != nil {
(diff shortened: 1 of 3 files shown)


hughhan1:
Updated this PR to keep credential-merge coverage at the integration/functional layer.

Changes in this follow-up:
- removed the config-level unit tests I had added earlier
- expanded the existing `TestListIntegrations_ConnectionInfosUseResolvedConnectionDefs` server test to assert field-level credential merging by name
- preserved coverage for both behaviors we care about: local field overrides on existing credentials and append behavior for local-only credentials

Validation:
- `go mod tidy && git diff --exit-code go.mod go.sum`
- `golangci-lint run`
- `go test ./...`
- `go test -coverprofile=coverage.out ./...`

GitHub checks on commit `fdc65817` are green, including `Go Tests`, `Go Coverage`, and `Go Lint & Vet`.

PR #588, opened 2026-04-06
Fix admin ECharts theme color resolution

hughhan1:
## Summary
- resolve admin chart theme colors through the browser before passing them to ECharts
- keep the fix local to the embedded admin page instead of changing shared theme tokens
- extend the existing admin Playwright flow to verify the page exposes ECharts-safe rgb/rgba colors

## Testing
- go test -ldflags='-w -s' ./cmd/gestaltd -run 'TestE2EInitServeLockedStdoutExposesPrometheusAndEmbeddedAdminUIByDefault'
- npm run check
- CI=1 npx playwright test e2e/auth.spec.ts

### DIFF
--- a/gestaltd/internal/adminui/out/index.html
+++ b/gestaltd/internal/adminui/out/index.html
@@ -576,20 +576,36 @@
           return charts[id];
         }
 
+        let themeProbe = null;
+
+        function resolveThemeColor(name) {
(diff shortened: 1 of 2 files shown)


hughhan1:
Superseded by #584, which now includes this fix.

PR #645, opened 2026-04-07
[codex] Publish gestaltd Docker images on release tags only

hughhan1:
## Summary

- remove the `main`/preview publish workflow for `gestaltd`
- make `gestaltd` Docker images release-tag only
- update the `gestaltd` Docker Hub doc to list only stable tags

## Why

We don't want preview images. `gestaltd` should publish Docker images only from release tags.

## Validation

- ran `git diff --check`


hughhan1:
Follow-up change pushed in `662830f2`: `gestaltd` preview builds now publish to `ghcr.io/valon-technologies/gestaltd-preview` instead of the main Docker Hub repo. Docker Hub is now stable-only for `gestaltd` (`latest` and `<version>`), while preview tags remain `:main` and `:<sha>` on GHCR.

Validation rerun after the change:
- parsed `.github/workflows/publish-gestaltd-image.yml` as YAML
- `git diff --check`
- `npm run build` in `docs/`

PR #657, opened 2026-04-07
Add Python SDK helper for explicit HTTP responses

hughhan1:
## Summary
- add `gestalt.respond(status, body)` alongside the existing `gestalt.OK(body)` helper
- re-export the helper from the public Python SDK surface
- cover non-200 helper behavior in the runtime tests

## Why
Python plugin handlers can already return `gestalt.Response(...)` for explicit non-200 responses, but the public API only had the `OK(...)` helper. This adds a symmetric helper for non-200 cases without introducing ambiguous tuple-return behavior.

### DIFF
--- a/sdk/python/gestalt/__init__.py
+++ b/sdk/python/gestalt/__init__.py
@@ -1,12 +1,34 @@
-from ._api import OK, Model, Request, Response, field
+from ._api import (
+    OK,
+    BadGateway,
+    BadRequest,
+    Conflict,
+    Forbidden,
+    InternalServerError,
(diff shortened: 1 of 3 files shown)


hughhan1:
Rebased onto current `main` and resolved the runtime test conflict. One API change from the original draft: I dropped the generic `respond(...)` helper in favor of named constructors like `gestalt.BadRequest(...)`, `gestalt.InternalServerError(...)`, `gestalt.NotFound(...)`, etc., since that was the stronger ergonomic direction.

PR #658, opened 2026-04-07
[codex] remove legacy gestaltd config path fallbacks

hughhan1:
## What changed

This removes the legacy `gestaltd` config auto-discovery paths and standardizes project-local server config on `.gestaltd/config.yaml`.

Specifically:
- removed bare `./config.yaml` auto-discovery
- removed legacy `~/.gestalt/config.yaml` and `/etc/gestalt/config.yaml` fallback support
- kept `GESTALT_CONFIG`, `./.gestaltd/config.yaml`, and `~/.gestaltd/config.yaml`
- adjusted default artifacts resolution so a config stored in `.gestaltd/config.yaml` does not create a nested `.gestaltd/.gestaltd/...` layout
- updated Docker, Helm, examples, and docs to use `/app/.gestaltd/config.yaml`
- added focused tests for lookup order and `.gestaltd` artifact layout

## Why

`./config.yaml` is too generic and can collide with other tools in the same repo. The old home/system paths were legacy artifacts that should no longer be documented or auto-supported.

## Impact

Project-local `gestaltd` usage now defaults to `.gestaltd/config.yaml`, which is safer and consistent with the prepared-state layout. Container and Helm examples now follow the same convention.

## Root cause

The server still carried historical path fallbacks in config resolution and deploy defaults, so the implementation and docs continued to teach and rely on legacy locations.

## Validation

- `gofmt -w cmd/gestaltd/factories.go cmd/gestaltd/factories_test.go internal/operator/localconfig.go internal/operator/lifecycle.go internal/operator/lifecycle_paths_test.go`
- `go test ./cmd/gestaltd ./internal/operator`


hughhan1:
Superseded by #660, which trims this down to only removing the implicit legacy lookup paths while keeping explicit /etc usage unchanged.

PR #865, opened 2026-04-12
fix: ad-hoc codesign darwin provider binaries

hughhan1:
## Summary

- Go's internal linker produces Mach-O ad-hoc signatures with the `linker-signed` flag (`0x20000`), which macOS 26 (Tahoe) rejects with SIGKILL
- Adds codesign step in `prepareBuiltPackageDir()` between binary compilation and SHA256 hashing
- Uses native `codesign` on macOS, falls back to `rcodesign` on Linux, warns and skips if neither is available

## Test plan

- [x] Verified cross-compiled binary changes from `flags=0x20002(adhoc,linker-signed)` to `flags=0x2(adhoc)` after signing
- [x] `gestaltd provider release --platform darwin/arm64` produces correctly signed binary
- [x] All existing plugin tests pass

### DIFF
--- a/gestaltd/internal/pluginstore/codesign.go
+++ b/gestaltd/internal/pluginstore/codesign.go
@@ -1,0 +1,22 @@
+package pluginstore
+
+import (
+	"fmt"
+	"os/exec"
+	"runtime"
+)
+
+// adhocCodesignDarwin ad-hoc signs a Mach-O binary on macOS to strip the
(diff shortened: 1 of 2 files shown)


hughhan1:
Closing — the signal: killed error is no longer reproducible. Downloaded darwin binaries with the linker-signed flag run fine on this macOS 26 machine, and gestaltd init completes natively. The root cause was likely transient (memory pressure, stale process, etc.), not codesign-related.

PR #881, opened 2026-04-12
Add IndexedDB cursor support to the Go SDK

hughhan1:
Implements the Cursor type and bidirectional streaming integration for the OpenCursor gRPC method. This adds OpenCursor and OpenKeyCursor on both ObjectStoreClient and IndexClient, along with cursor navigation (Continue, ContinueToKey, Advance), value access (Key, PrimaryKey, Value), and mutation (Delete, Update) methods.


hughhan1:
Folded into #878

PR #882, opened 2026-04-12
Add IndexedDB cursor support to the Rust SDK

hughhan1:
Update the existing conversion helpers to work with the new TypedValue/Record proto messages instead of prost_types::Struct/Value. Add a Cursor struct backed by a bidirectional gRPC stream (mpsc sender for commands, tonic::Streaming for responses) with methods for iteration, seeking, in-place mutation, and cleanup. Expose open_cursor and open_key_cursor on both ObjectStore and IndexClient, and export Cursor and CursorDirection from the crate root.


hughhan1:
Folded into #878

PR #883, opened 2026-04-12
Add cursor support to TypeScript SDK

hughhan1:
Adds bidi-streaming cursor iteration to the IndexedDB TypeScript SDK. This introduces an AsyncQueue class for push-based request feeding into the Connect-RPC bidi stream, a Cursor class with open/continue/advance/delete/ update/close operations, and a CursorDirection enum. Both ObjectStore and Index classes now expose openCursor and openKeyCursor methods. The generated proto stubs are regenerated to include the new cursor message types and the OpenCursor streaming RPC.


hughhan1:
Folded into #878

PR #884, opened 2026-04-12
Add cursor support to Python SDK for IndexedDB

hughhan1:
Implements bidirectional streaming cursor iteration using the OpenCursor gRPC method. The Cursor class wraps the bidi stream with a queue-based request iterator, supporting continue_, continue_to_key, advance, delete, update, and close operations, along with key/primary_key/value properties. ObjectStore and Index both gain open_cursor and open_key_cursor methods. Also exports CURSOR_NEXT, CURSOR_NEXT_UNIQUE, CURSOR_PREV, and CURSOR_PREV_UNIQUE direction constants, and fixes _to_proto_value -> _to_typed_value in Index._req.


hughhan1:
Folded into #878

PR #941, opened 2026-04-14
Prepare Homebrew formulas for public release assets

hughhan1:
## Summary
- remove the custom Homebrew download strategy used for private GitHub release assets
- delete the extra Ruby helper under `Formula/lib/`
- switch both formulas back to standard GitHub release URLs

## Important
This change is not safe to merge while `valon-technologies/gestalt` is still private.

As of April 15, 2026, the repository is still private, so anonymous requests to:

`https://github.com/valon-technologies/gestalt/releases/download/...`

return `404`. That means `GitHubPrivateRepositoryReleaseDownloadStrategy` is still required for working tap installs today.

This PR should stay draft until one of these is true:
- the repository becomes public
- the release artifacts move to a public host

## Current vs Future Install Behavior
Current private-repo behavior:
- with the private strategy: `brew install` can authenticate to release assets
- without the private strategy: GitHub returns `404` for release downloads

Future public-repo/public-asset behavior:
- the private strategy becomes unnecessary maintenance-only code
- standard Homebrew release URLs are sufficient

## Validation
- `ruby -c Formula/gestalt.rb`
- `ruby -c Formula/gestaltd.rb`
- `rg -n "private_strategy|GitHubPrivateRepositoryReleaseDownloadStrategy" Formula`


hughhan1:
Follow-up from an independent review pass: this diff is correct once release assets are publicly downloadable, but it will regress `brew install`/`brew upgrade` while the repository releases are still auth-gated. Keeping this PR in draft until the repo/release visibility flip happens avoids introducing that regression prematurely.

PR #981, opened 2026-04-14
Point cache docs at cache/valkey

hughhan1:
## Summary

Replace the remaining placeholder `cache/redis` docs examples with the real first-party `cache/valkey` provider shape.

This PR is docs-only and depends on the provider repo rollout at `valon-technologies/gestalt-providers#105`.

### YAML

```yaml
providers:
  cache:
    session:
      source:
        path: /path/to/local/cache-provider/manifest.yaml
      config:
        addresses:
          - ${VALKEY_ADDR}
        db: 0
        username: ${VALKEY_USERNAME}
        password: secret://valkey-password
        tls: false
        dialTimeout: 5s
        writeTimeout: 5s

plugins:
  search:
    cache:
      - session
```

Once `cache/valkey` is published as a managed package, that same provider can be referenced as:

```yaml
source:
  ref: github.com/valon-technologies/gestalt-providers/cache/valkey
  version: 0.0.1-alpha.1
```

### Go

```go
cache, err := gestalt.Cache("session")
if err != nil {
    panic(err)
}
defer cache.Close()

value, found, err := cache.Get(ctx, "user:1")
if err != nil {
    panic(err)
}
_ = value
_ = found
```

## Updated Docs

- `docs/content/providers/cache.mdx`
- `docs/content/reference/config-file.mdx`
- `docs/content/custom-providers/cache.mdx`
- `docs/content/providers/index.mdx`

The docs now:

- remove the old `cache/redis` placeholder references
- use the real `cache/valkey` config shape (`addresses`, `db`, `username`, `password`, `tls`, `dialTimeout`, `writeTimeout`)
- use local `source.path` examples until the managed package is published
- keep the custom-provider manifest example generic instead of pretending to be the exact first-party Valkey manifest

## Testing

Docs-only change. I verified the docs content no longer contains `cache/redis`, `REDIS_ADDR`, or `Redis Cache` references.

### DIFF
--- a/docs/content/custom-providers/cache.mdx
+++ b/docs/content/custom-providers/cache.mdx
@@ -16,10 +16,10 @@ A cache provider manifest declares `kind: cache`:
 
 ```yaml
 kind: cache
-source: github.com/valon-technologies/gestalt-providers/cache/redis
-version: 0.0.1
-displayName: Redis Cache
-description: Cache provider backed by Redis or Valkey.
+source: github.com/your-org/gestalt-providers/cache/customcache
+version: 0.0.1-alpha.1
+displayName: Custom Cache
+description: Example custom cache provider.
 spec:
   configSchemaPath: ./cache_config.json
 ```
@@ -41,7 +41,7 @@ The Cache service includes:
 <Tabs items={['Go', 'Python', 'Rust', 'TypeScript']}>
   <Tabs.Tab>
     ```go
-    package rediscache
+    package customcache
 
     import (
         "context"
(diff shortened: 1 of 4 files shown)


hughhan1:
Corrected interface note after review: the docs branch now uses a local `source.path` Valkey manifest in the primary examples because `cache/valkey/v0.0.1-alpha.1` has not been published yet.

```yaml
providers:
  cache:
    session:
      source:
        path: /path/to/gestalt-providers/cache/valkey/manifest.yaml
      config:
        addresses:
          - ${VALKEY_ADDR}
        db: 0
        username: ${VALKEY_USERNAME}
        password: secret://valkey-password
        tls: false
        dialTimeout: 5s
        writeTimeout: 5s
```

Once the managed package is published, that same provider can be referenced as:

```yaml
source:
  ref: github.com/valon-technologies/gestalt-providers/cache/valkey
  version: 0.0.1-alpha.1
```

The custom-provider manifest example was also made generic again so it no longer implies it is the exact first-party Valkey manifest.

PR #1026, opened 2026-04-15
gestaltd: add executable plugin invocation runtime

hughhan1:
## Summary
- add the host-issued request-handle snapshot path for executable plugin-to-plugin invocation
- expose a per-plugin host `PluginInvoker` service gated by `config.yaml` `invokes` declarations
- add the Go SDK request-scoped invoker helper and bootstrap functional coverage for nested calls

## New interfaces
YAML:
```yaml
plugins:
  caller:
    invokes:
      - plugin: example
        operation: request_context
```

Go:
```go
func (r Request) Invoker() (*InvokerClient, error)
func (c *InvokerClient) Invoke(ctx context.Context, plugin, operation string, params map[string]any, opts *InvokeOptions) (*OperationResult, error)

type InvokeOptions struct {
    Connection string
    Instance   string
}
```

Example:
```go
invoker, err := req.Invoker()
if err != nil {
    return gestalt.Response[Out]{}, err
}
defer invoker.Close()

result, err := invoker.Invoke(ctx, "example", "request_context", map[string]any{
    "ticket_id": "123",
}, &gestalt.InvokeOptions{
    Connection: "work",
})
if err != nil {
    return gestalt.Response[Out]{}, err
}
```

## Implementation notes
- `ExecuteRequest.invocation_id` now carries an opaque host-issued request handle for invokes-enabled executable plugins
- the host stores the authoritative principal/access/depth/call-chain/ambient-connection snapshot and restores it for nested calls instead of trusting plugin-supplied request context
- omitted nested `connection` inherits the outer ambient connection; explicit nested `connection` overrides it
- raw executable providers served through `providerhost` now receive the request handle in `context.Context` for internal host-service access

## Tests
- `go test ./internal/providerhost`
- `go test ./internal/bootstrap`
- `go test ./...` in `sdk/go`
- `go test ./internal/testplugins/echo`
- `go test ./...` in `gestaltd/internal/testutil/testdata/provider-go`


hughhan1:
Updated interface note after review fixes:

YAML remains:
```yaml
plugins:
  caller:
    invokes:
      - plugin: example
        operation: request_context
```

Go now exposes both request-scoped paths:
```go
func (r Request) Invoker() (*InvokerClient, error)
func InvokerFromContext(ctx context.Context) (*InvokerClient, error)
func (c *InvokerClient) Invoke(ctx context.Context, plugin, operation string, params map[string]any, opts *InvokeOptions) (*OperationResult, error)
```

Proto now carries a dedicated nested-call capability token without repurposing `invocation_id`:
```proto
message ExecuteRequest {
  string invocation_id = 5;
  RequestContext context = 6;
  string request_handle = 7;
}
```

I also added functional coverage for undeclared target rejection and invalid/expired request-handle rejection in `gestaltd/internal/bootstrap/executable_plugin_test.go`.

PR #1029, opened 2026-04-15
[codex] Fix Homebrew install instructions

hughhan1:
## What changed
- add the explicit `brew tap valon-technologies/gestalt https://github.com/valon-technologies/gestalt` step to the top-level install docs
- update the Homebrew install page to explain why the extra tap command is required
- keep the existing install commands for `gestaltd` and `gestalt` after the tap is configured

## Why
Homebrew's default `user/repo` tap shortcut assumes a repository named `homebrew-gestalt`. That repository currently does not exist, so `brew install valon-technologies/gestalt/gestalt` prompts for GitHub credentials and then fails while trying to clone a 404 tap.

## User impact
Users following the docs will get a working install path again on a fresh machine.

This is a short-term documentation fix. The cleaner long-term Homebrew setup would be either:
- restore a dedicated `valon-technologies/homebrew-gestalt` tap, or
- publish through `homebrew/core`

## Validation
- `git diff --check`
- manually verified that `https://github.com/valon-technologies/homebrew-gestalt` currently returns 404 and that the main `valon-technologies/gestalt` repo and public release assets are reachable

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,9 +30,10 @@ Gestalt also does not replace your existing APIs. It sits between agents and ups
 
 ## Quick Start
 
-Install both binaries with Homebrew:
+Tap the repository and install both binaries with Homebrew:
 
 ```sh
+brew tap valon-technologies/gestalt https://github.com/valon-technologies/gestalt
 brew install valon-technologies/gestalt/gestaltd
 brew install valon-technologies/gestalt/gestalt
 ```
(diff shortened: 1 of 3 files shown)


hughhan1:
Superseded by restoring the dedicated Homebrew tap repository instead of documenting the custom-remote workaround.

PR #1041, opened 2026-04-16
server: dedupe mounted auth coverage and cover refresh token URL

hughhan1:
## Summary
This PR trims redundant server-level test surface in `gestaltd/internal/server/server_test.go` and adds one missing functional test for the resolved `RefreshTokenWithURL(...)` branch.

Changes in this slice:
- rename the existing default-allow mounted UI case so it explicitly covers the `PluginName` branch
- keep a separate no-`PluginName` fallback case instead of duplicating the plugin-name fixture
- add one server-level refresh test that proves resolved connection params drive `RefreshTokenWithURL(...)`

## Go Interface Changes
- None

### Go Example
```go
// No Go interface changes in this PR.
```

## YAML Interface Changes
- None

### YAML Example
```yaml
# No YAML interface changes in this PR.
```

## Verification
- `cd /Users/hugh/src/gestalt/gestaltd && go test ./internal/server -count=1`
- `cd /Users/hugh/src/gestalt/gestaltd && go test ./internal/... -run TestDoesNotExist -count=1`

## Notes
- This is PR 1 from the internal reduction plan: baseline repair plus one missing product-level server test.
- No unit tests were added.

### DIFF
--- a/gestaltd/internal/server/handlers.go
+++ b/gestaltd/internal/server/handlers.go
@@ -505,16 +505,7 @@ func (s *Server) listOperations(w http.ResponseWriter, r *http.Request) {
 	var cat *catalog.Catalog
 	if strictCatalog && requestedConnection == "" && requestedInstance == "" &&
 		(s.authorizer == nil || !s.authorizer.IsWorkload(p)) {
-		catalogConnections := make([]string, 0, 2)
-		if conn := s.catalogLookupConnection(name, requestedConnection); conn != "" {
-			catalogConnections = append(catalogConnections, conn)
-		}
-		if conn := s.defaultConnection[name]; conn != "" && (len(catalogConnections) == 0 || catalogConnections[0] != conn) && s.catalogConnection[name] == "" {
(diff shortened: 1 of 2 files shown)


hughhan1:
Updated scope for this PR after the follow-on reduction slice:

- extracted `sessionCatalogConnections(...)` so `listOperations` and `executeOperation` use the same session-catalog connection precedence/fallback logic
- kept the mounted WebUI default-allow coverage split into explicit `PluginName` and no-`PluginName` cases
- consolidated the refresh-path server tests into broader product-behavior cases covering:
  - pass-through when refresh does not apply
  - persisted token field updates after refresh
  - expired-token hard failure and still-valid/deleted-token fallback
  - direct `RefreshToken(...)` and resolved `RefreshTokenWithURL(...)` paths

Go interface changes: none
YAML interface changes: none
Examples: N/A

Verification rerun on this updated branch:
- `cd gestaltd && go test ./internal/server -count=1`
- `cd gestaltd && go test ./internal/... -run TestDoesNotExist -count=1`

No unit tests were added.

PR #1068, opened 2026-04-16
gestaltd: share static connection plan

hughhan1:
## Summary
- move manifest-backed connection selection into a shared `internal/config` planner so bootstrap runtime, resolved-structure validation, and prepared-provider validation all consume the same logic
- remove duplicated bootstrap-only branching for connection resolution and prepared-provider connection-mode aggregation
- validate undeclared `rest/openapi/graphql/mcp` surface connection refs during resolved-structure load using the same planner
- keep connection advertisement and runtime connection selection behavior unchanged while reducing the number of separate decision paths

## Go Interface Changes
- added internal config API: `BuildStaticConnectionPlan(*ProviderEntry, *providermanifestv1.Spec) (StaticConnectionPlan, error)`
- added internal config types for shared consumers: `StaticConnectionPlan` and `ResolvedSpecSurface`
- example:

```go
plan, err := config.BuildStaticConnectionPlan(entry, entry.ManifestSpec())
if err != nil {
    return err
}

apiConn := plan.APIConnection()
mode := plan.ConnectionMode()
resolved, ok := plan.ConfiguredSpecSurface()
```

## YAML Interface Changes
- none; existing connection fields are unchanged
- the same fields now validate against the shared planner during resolved-structure load
- example unchanged YAML:

```yaml
plugins:
  example:
    defaultConnection: plugin
    source:
      path: ./manifest.yaml
```

## Verification
- `cd gestaltd && go test ./internal/config ./internal/bootstrap ./internal/operator -count=1`
- `cd gestaltd && go test ./internal/server ./internal/mcp -count=1`
- `cd gestaltd && go test ./internal/... -run TestDoesNotExist -count=1`

## Test Coverage
- adds bootstrap coverage for manifest-aware prepared-provider connection mode
- adds operator execution-path coverage for undeclared `rest/openapi/graphql/mcp` surface connection refs
- adds no new unit-test-only helper suite

### DIFF
--- a/gestaltd/internal/authorization/authorizer.go
+++ b/gestaltd/internal/authorization/authorizer.go
@@ -18,7 +18,6 @@ import (
 	"github.com/valon-technologies/gestalt/server/internal/emailutil"
 	"github.com/valon-technologies/gestalt/server/internal/principal"
 	"github.com/valon-technologies/gestalt/server/internal/registry"
-	providermanifestv1 "github.com/valon-technologies/gestalt/server/sdk/providermanifest/v1"
 )
 
 const (
@@ -816,7 +815,11 @@ func normalizeEmail(email string) string {
 
 func providerMode(provider string, pluginDefs map[string]*config.ProviderEntry, providers *registry.ProviderMap[core.Provider]) (core.ConnectionMode, bool, error) {
 	if entry, ok := pluginDefs[provider]; ok && entry != nil {
-		return pluginConnectionMode(entry), true, nil
+		plan, err := config.BuildStaticConnectionPlan(entry, entry.ManifestSpec())
+		if err != nil {
+			return "", false, err
+		}
+		return plan.ConnectionMode(), true, nil
 	}
 	if providers != nil {
 		prov, err := providers.Get(provider)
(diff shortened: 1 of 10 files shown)


hughhan1:
Reworked this PR to meet the stricter reduction bar.

Changes since the first version:
- rebased the branch onto current `origin/main` so the PR no longer drags unrelated SDK/runtime history
- pushed the shared static connection plan farther through the stack so authz and server connection lookup also consume it instead of re-deriving connection mode / plugin-vs-named connection state
- removed the extra bootstrap helper layer for advertised connection names
- switched MCP OAuth spec-provider build options to use the resolved MCP surface URL from the shared plan instead of bypassing it through the raw manifest lookup

Current diff against `main`:
- production code: `+423 / -567` (net `-144`)
- total including tests: `+550 / -567` (net `-17`)

Verification rerun after the rework:
- `cd gestaltd && go test ./internal/config ./internal/bootstrap ./internal/operator ./internal/authorization ./internal/server ./internal/mcp -count=1`
- `cd gestaltd && go test ./internal/... -run TestDoesNotExist -count=1`

A fresh `gpt-5.4 xhigh` review pass on the updated branch found no correctness issues.

PR #1070, opened 2026-04-16
sdk/typescript: trim duplicate s3 behavior coverage

hughhan1:
## Summary
- keep `tests/s3.transport.test.ts` as the canonical S3 behavior suite
- remove overlapping S3 write-stream assertions from `tests/runtime.test.ts` while preserving direct runtime-adapter coverage for `headObject`, `listObjects`, `copyObject`, and `presignObject`
- keep the packaged S3 build smoke lean, but preserve the post-write `headObject` assertion so the compiled binary proves a second unary RPC against persisted state

## Interfaces
Go
- No Go interface change in this slice.

YAML
```yaml
provider:
  target: ./provider.ts#plugin
```

TypeScript
```ts
const s3 = new S3();
await s3.object("build-bucket", "hello.txt").writeString("hello");
```

## Verification
- `bun test tests/s3.transport.test.ts tests/runtime.test.ts tests/build.test.ts`
- `bun run check`

## Stack
- Stacked on top of #1069

### DIFF
--- a/sdk/typescript/tests/build.test.ts
+++ b/sdk/typescript/tests/build.test.ts
@@ -746,6 +746,7 @@ test("buildProviderBinary compiles a runnable s3 provider executable", async () 
       },
     });
     expect(headed.meta?.size).toBe(5n);
+
   } finally {
     if (child) {
       await stopProcess(child);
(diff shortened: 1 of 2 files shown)


hughhan1:
Dropping this slice. After restoring the adapter-path coverage that reviewers correctly called out, the diff collapsed into near-zero net reduction and isn’t worth carrying as a standalone PR. The next real cleanup continues in the stacked code slice instead.

PR #1071, opened 2026-04-16
sdk/typescript: normalize s3 body inputs

hughhan1:
## Summary
- share one byte-backed S3 conversion primitive between the stream path and the write-time snapshot path
- snapshot mutable `Uint8Array`/`ArrayBuffer`/view inputs once in `S3.writeObject()` using the existing `S3BodySource` union
- remove duplicate byte-shape branching without changing the public S3 API

## Interfaces
Go
- No Go interface change in this slice.

YAML
```yaml
provider:
  target: ./provider.ts#plugin
```

TypeScript
```ts
const s3 = new S3();
await s3.object("docs-bucket", "payload.bin").writeBytes(new Uint8Array([1, 2, 3]));
```

## Verification
- `bun test tests/s3.transport.test.ts tests/runtime.test.ts tests/build.test.ts`
- `bun run check`

## Stack
- Stacked on top of #1069

### DIFF
--- a/sdk/typescript/src/plugin.ts
+++ b/sdk/typescript/src/plugin.ts
@@ -64,6 +64,15 @@ export interface OperationDefinition<In, Out> extends OperationOptions<
   Out
 > {}
 
+type StoredOperationDefinition<In, Out> = OperationDefinition<In, Out> & {
+  id: string;
+  method: string;
+  title: string;
+  description: string;
+  allowedRoles: string[];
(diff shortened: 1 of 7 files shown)


hughhan1:
Closing this as subsumed by #1069. I rewrote #1069 to carry the net-negative S3 production-code cleanup directly, so keeping this stacked follow-up open would just duplicate the same slice.

PR #1072, opened 2026-04-16
core: preserve sparse session catalog metadata

hughhan1:
## Summary
Preserve static REST metadata when a session catalog publishes a sparse operation with the same ID, while also collapsing the split strict/non-strict list-operations resolution path into one code path.

Non-test diff: `+117 / -120` (net `-3`)
Test diff: `+119 / -25` (net `+94`)

## Go Interface
```go
cat, meta, err := invocation.ResolveCatalogWithMetadata(
    ctx,
    prov,
    "sample-int",
    resolver,
    p,
    "catalog-conn",
    "",
    true,
)
```

Behavior:
- `strictSession=true` keeps explicit/session-backed discovery strict
- session `method` / `description` / `allowedRoles` stay authoritative on same-ID collisions
- static `path` / `query` / parameter `wireName` + `location` backfill sparse session collisions

## YAML Interface
```yaml
# static catalog
operations:
  - id: alpha_rest
    method: GET
    path: /items/{id}
    query: query StaticItems { items { id } }
    parameters:
      - name: itemId
        wireName: id
        location: path
      - name: pageSize
        wireName: page[size]
        location: query

# session catalog
operations:
  - id: alpha_rest
    method: POST
    description: Session REST op
    parameters:
      - name: itemId
      - name: pageSize
```

Merged result:
```yaml
operations:
  - id: alpha_rest
    method: POST
    description: Session REST op
    path: /items/{id}
    query: query StaticItems { items { id } }
    parameters:
      - name: itemId
        wireName: id
        location: path
      - name: pageSize
        wireName: page[size]
        location: query
```

## Verification
```bash
go test ./internal/server ./internal/invocation ./internal/mcp
golangci-lint run ./internal/server ./internal/invocation ./internal/mcp
```


hughhan1:
Follow-up after reviewer feedback on head `aa79caeb`.

Updated diff on the current branch:
- non-test: `+201 / -207` (net `-6`)
- test: `+535 / -25` (net `+510`)

Applied changes since the previous push:
- `core/integration.Base` now executes from the resolved catalog operation context, and only takes the GraphQL path for GraphQL-shaped operations instead of any op that merely carries a `query`
- same-ID session overrides no longer inherit a stale static transport when the session op already declares a REST-shaped `method` / `path`
- MCP dynamic session tools now hydrate from the same merged session/static operation metadata they execute with, so sparse same-ID collisions expose the required path/query parameter schema instead of a stale sparse tool shape
- aliased `core/integration.Restricted` wrappers now reseed the resolved operation context under the inner operation name before delegating, so merged session metadata reaches inner `Base` providers

Go:
```go
cat, meta, err := invocation.ResolveCatalogWithMetadata(
    ctx,
    prov,
    "sample-int",
    resolver,
    p,
    "catalog-conn",
    "",
    true,
)

op, transport, connection, err := invocation.ResolveOperation(
    ctx,
    prov,
    "sample-int",
    resolver,
    p,
    "run",
    []string{"catalog-conn"},
    "",
)
```

YAML merged result:
```yaml
operations:
  - id: run
    method: POST
    description: Session REST op
    path: /items/{itemId}/run
    parameters:
      - name: pageSize
        wireName: page[size]
        location: query
      - name: itemId
        wireName: itemId
        location: path
        required: true
```

Verification rerun on the updated branch:
```bash
go test ./internal/server ./internal/invocation ./internal/mcp ./core/integration
golangci-lint run ./internal/server ./internal/invocation ./internal/mcp ./core/integration
```

PR #1078, opened 2026-04-16
core: preserve hydrated session operation metadata

hughhan1:
Go interface:
- invocation.ResolveCatalogWithMetadata(...) now returns (*catalog.Catalog, bool, error), where the bool reports only whether session resolution failed.
- invocation.HydrateSessionCatalog(staticCat, sessionCat) hydrates sparse session catalogs before discovery and MCP projection.

Go example:
  cat, sessionFailed, err := invocation.ResolveCatalogWithMetadata(ctx, prov, "sampledb", resolver, p, "catalog-conn", "", false)
  if sessionFailed {
      slog.Warn("falling back to static catalog")
  }

YAML interface:
  Static catalogs still declare reusable transport and parameter metadata, while session catalogs can override request-scoped behavior for the same operation id.

YAML example static catalog:
  name: sampledb
  operations:
    - id: run_query
      method: POST
      parameters:
        - name: sql
          type: string
          required: true

YAML example session catalog:
  name: sampledb
  operations:
    - id: run_query
      description: session query
      transport: rest

Behavior:
- session discovery and MCP tool hydration now use the same hydrated session operation shape
- sparse session operations keep static transport, path, headers, and parameter wiring
- the unused SessionAttempted metadata branch is removed without changing discovery order

### DIFF
--- a/gestaltd/internal/invocation/catalog.go
+++ b/gestaltd/internal/invocation/catalog.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"maps"
 
 	"github.com/valon-technologies/gestalt/server/core"
 	"github.com/valon-technologies/gestalt/server/core/catalog"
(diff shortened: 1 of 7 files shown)


hughhan1:
Updated the branch to a single rewritten commit: `06796d7e` (`core: preserve hydrated session operation metadata`).

The current interface shape on the branch is:

Go:
```go
cat, attempted, err := core.CatalogForRequest(ctx, prov, token)
if attempted && err != nil {
    return err
}

cat, sessionFailed, err := invocation.ResolveCatalogWithMetadata(
    ctx,
    prov,
    "sampledb",
    resolver,
    p,
    "catalog-conn",
    "",
    false,
)
if sessionFailed {
    slog.Warn("falling back to static catalog")
}
```

YAML:
```yaml
name: sampledb
operations:
  - id: run_query
    method: POST
    parameters:
      - name: sql
        type: string
        required: true
```

```yaml
name: sampledb
operations:
  - id: run_query
    description: session query
    transport: rest
```

I attempted to update the PR title/body directly, but `gh pr edit` is currently blocked by GitHub API rate limiting on the authenticated user. I’ll retry later if the rate limit clears.

PR #1086, opened 2026-04-16
gestaltd: collapse config secret resolution helpers

hughhan1:
## Summary
This PR removes one full layer of config-secret-resolution indirection in `gestaltd/internal`.

It does three things:
- collapses bootstrap secret resolution into a single `ResolveConfigSecrets(...)` path and deletes the separate `internal/bootstrap/secret_resolution.go` helper
- inlines the compatibility-only canonicalization wrapper into `CanonicalizeStructure(...)`
- removes the extra operator helper layer for secrets-provider metadata dependency/resolution, folding that work directly into `primeSecretsProviderForConfigResolution(...)`

The result is less production code, fewer helper hops, and less branching in the config-secret bootstrap flow without changing the external product surface.

## Interfaces
Go interface: none
YAML interface: none

## Examples
Go example (unchanged):
```go
if err := bootstrap.ResolveConfigSecrets(ctx, cfg, factories); err != nil {
    return err
}
```

YAML example (unchanged):
```yaml
server:
  encryptionKey:
    secret:
      provider: default
      name: api-token
```

## Verification
- `go test ./internal/bootstrap ./internal/operator ./internal/config -count=1`
- `go test ./internal/operator -run 'TestSourceAuthPluginLoadForExecution|TestSourceSecretsPluginBootstrapsManagedAuthSourceToken|TestInitAtPath_RejectsDisabledManagedHostProvidersWithStructuredSecretRefs' -count=1`
- `go test ./internal/... -run TestDoesNotExist -count=1`

## Diff
Production code vs `main`: `+65 / -118` (net `-53` LOC)


hughhan1:
Addressed the post-resolution canonicalization review note in `ResolveConfigSecrets(...)`.

Change:
- restore `config.CanonicalizeStructure(cfg)` after `config.TransformConfigStringFields(...)` so the resolver still leaves the config in canonicalized form before the lifecycle boundary calls `ValidateCanonicalStructure(...)`

Re-ran:
- `go test ./internal/bootstrap ./internal/operator ./internal/config -count=1`
- `go test ./internal/operator -run 'TestSourceAuthPluginLoadForExecution|TestSourceSecretsPluginBootstrapsManagedAuthSourceToken|TestInitAtPath_RejectsDisabledManagedHostProvidersWithStructuredSecretRefs' -count=1`
- `go test ./internal/bootstrap -run TestBootstrapSecretResolution -count=1`

I also re-ran `go test ./internal/... -run TestDoesNotExist -count=1`; a separate broad package run had an unrelated bootstrap flake in `TestPythonSourcePluginFallsBackWithoutGoOnPath`, but the targeted bootstrap/operator/config verification for this change passed.

PR #1162, opened 2026-04-20
treat headless identity callers as server clients without user context

hughhan1:
## Summary

This PR treats any caller without user context as a headless server client for route and UI gating.

## Old interfaces

Go:

```go
if p != nil && p.IsWorkload() {
    // hide connect/settings affordances and reject user-only routes
}

if p == nil || p.IsWorkload() || strings.TrimSpace(p.UserID) == "" {
    return limitedSurface
}
```

Example:

```go
if p != nil && p.IsWorkload() {
    info.MountedPath = s.integrationMountedPathForPrincipalContext(ctx, p, name, info.MountedPath)
    continue
}
```

YAML:

```yaml
# no public config changes in this PR
```

Behavior:
- route/UI gating treated workload kind as the only headless case.
- empty-`UserID` checks were scattered through user-facing handlers.

## New interfaces

Go:

```go
func (p *Principal) HasUserContext() bool
```

Example:

```go
if p != nil && !p.HasUserContext() {
    // hide connect/settings affordances and reject user-only routes
}
```

YAML:

```yaml
# no public config changes in this PR
```

Behavior:
- user-only routes and mounted UI gating now key off `HasUserContext()`.
- authenticated headless identity callers are blocked from user-only surfaces.
- policyless mounted UIs stay public when no auth is presented.
- integration listing keeps headless callers on the limited server-client surface.

### DIFF
--- a/gestaltd/internal/principal/principal.go
+++ b/gestaltd/internal/principal/principal.go
@@ -66,6 +66,20 @@ func (p *Principal) AuthSource() string {
 	return p.Source.String()
 }
 
+func (p *Principal) IsWorkload() bool {
+	return p != nil && p.Kind == KindWorkload
+}
+
+func (p *Principal) HasUserContext() bool {
+	if p == nil {
+		return false
+	}
+	if strings.TrimSpace(p.UserID) != "" {
+		return true
+	}
+	return p.Identity != nil && strings.TrimSpace(p.Identity.Email) != ""
+}
+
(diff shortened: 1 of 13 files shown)


hughhan1:
Closing this stacked transitional PR. I am rebuilding the workload-removal change directly on fresh origin/main instead of continuing the old additive stack.

PR #1538, opened 2026-04-29
Close IndexedDB cursor after update transport check

hughhan1:
## Summary
- Closes the cursor stream in the TypeScript IndexedDB cursor-update transport case after validating the mutation.
- Keeps the existing W3C IndexedDB API contract unchanged; this is a test lifecycle cleanup only.

## Interfaces
None.

## Example
```ts
await cursor.update({ id: "u1", status: "inactive" });
expect(cursor.value).toEqual({ id: "u1", status: "inactive" });
cursor.close();
```

## Validation
- `bun test tests/indexeddb.transport.test.ts && bun run typecheck` from `sdk/typescript`


hughhan1:
Closing as redundant: the cursor close cleanup is already present on current main via #1542, and the idempotency work is continuing in #1536.

PR #1567, opened 2026-04-30
Recover accepted agent turns after create timeout

hughhan1:
## Problem

Slack-triggered workflows can call `Manager.CreateTurn` through a hosted agent provider. If the provider accepts and persists the deterministic turn but the provider RPC returns `DeadlineExceeded`, the workflow currently fails before it can observe the accepted turn. That leaves a valid turn running without the workflow waiting on it.

## Interface

No new public API is required. Callers keep using `Manager.CreateTurn` with a deterministic idempotency key:

```go
turn, err := manager.CreateTurn(ctx, principal, coreagent.ManagerCreateTurnRequest{
    SessionID: session.ID,
    IdempotencyKey: "workflow:slack:run-123:turn:signal-batch",
    Messages: []coreagent.Message{{Role: "user", Text: "..."}},
    CallerPluginName: "slack",
})
```

Providers keep the existing contract: if `CreateTurn` accepted the request before the deadline, `GetTurn` for the deterministic turn ID should return that turn.

## Logic Flow

1. `Manager.CreateTurn` computes the deterministic turn ID from `SessionID + IdempotencyKey`.
2. It calls the provider `CreateTurn` as before.
3. If the provider returns success, behavior is unchanged.
4. If the provider returns an error, the existing immediate `GetTurn` fallback still runs.
5. If the create error is retryable at the provider boundary (`DeadlineExceeded` / `Unavailable`) and `GetTurn` initially misses or hits a transient read error, the manager briefly polls `GetTurn` for the same turn ID.
6. If the turn appears, the manager normalizes and authorizes it exactly like a direct create response. If it does not appear, the original create error is preserved.

## Verification

- `go test -count=1 ./internal/agentmanager`
- `go test -count=1 ./internal/bootstrap -run 'TestHostedAgentProviderPool(GetTurnRetriesAfterPreferredTimeout|GetTurnRetriesAfterStalePreferredMiss|SkipsPastDrainBackendForNewTurn)'`\n\nReviewer agent: no findings; also ran `go test -count=1 ./internal/agentmanager`.

### DIFF
--- a/gestaltd/internal/agentmanager/manager.go
+++ b/gestaltd/internal/agentmanager/manager.go
@@ -53,6 +53,8 @@ const (
 	maxAgentRouteCacheEntries         = 20_000
 	AgentListSummaryDefaultLimit      = 100
 	AgentListMaxLimit                 = 500
+	agentCreateTurnRecoveryTimeout    = 15 * time.Second
+	agentCreateTurnRecoveryInterval   = 250 * time.Millisecond
 )
 
 type AgentProviderNotAvailableError struct {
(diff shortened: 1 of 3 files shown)


hughhan1:
Parking this PR for now. The cleaner follow-up is https://github.com/valon-technologies/gestalt-providers/pull/359, which moves the retry/recovery boundary into `workflow/indexeddb` by retrying the claimed agent signal batch durably instead of adding a one-off `CreateTurn` polling window in the agent manager.

hughhan1:
Closing this parked tactical recovery PR now that the cleaner workflow/indexeddb durable retry replacement merged in valon-technologies/gestalt-providers#359. The long-term path is to deploy that provider release instead of adding host-side DeadlineExceeded polling.

PR #1644, opened 2026-05-01
[codex] Harden operation and OAuth URL contracts

hughhan1:
## Summary
- Escape REST path-template parameter values as path segments so values such as `../../admin/users` cannot traverse to a different upstream path.
- Require restricted GraphQL calls to name an allowed catalog operation, then execute that catalog operation instead of arbitrary caller-supplied GraphQL documents.
- Apply the same allowed-operation GraphQL rule to session-backed GraphQL catalog providers.
- Split provider metadata validation from token-response connection-param validation so token-derived URL template params reject `:` and `/`.
- Validate connection params before OAuth refresh URL interpolation to block persisted unsafe values from changing token endpoint authority/path.

## Verification
- `go test ./services/plugins/apiexec ./services/plugins/declarative ./core/integration ./services/invocation -run 'TestSubstitutePath|TestBaseExecuteEscapesPathParameterSegments|TestRestrictedInvokeGraphQLRequiresAllowedCatalogOperation|TestBrokerResolveToken_ConnectionModeNoneResolvesSessionUserSubject' -count=1`
- `go test ./internal/bootstrap -run 'TestGraphQLSessionCatalogProviderLoadsCatalogOnDemand|TestGraphQLSessionCatalogProviderInvokeGraphQLRequiresAllowedOperation' -count=1`
- `go test ./internal/server -run 'TestIntegrationOAuthCallback|TestRefresh_UsesConnectionAuthHandlers' -count=1`
- `go test ./services/plugins/apiexec ./services/plugins/declarative ./core/integration ./services/invocation ./internal/bootstrap ./internal/server`
- `git diff --check`


hughhan1:
Rebased again onto current `origin/main` (`a50218bcfe7dff8ca5cfc2c6205cc26d4089ad29`) and force-pushed.

Adapted the branch to the latest service refactors by moving broker tests to `services/authorization` and updating the GraphQL session provider test to use `declarative.Build`.

Verification after this rebase:
- `go test ./services/plugins/apiexec ./services/plugins/declarative ./services/invocation ./internal/bootstrap ./internal/server ./services/plugininvoker ./services/plugins/operationexposure`
- `go test ./...` in `sdk/go`
- `uv run --project sdk/python pytest sdk/python/tests/test_plugin_invoker_transport.py`
- `cargo test --test plugin_invoker_transport` in `sdk/rust`
- `bun run typecheck && bun test --max-concurrency 1` in `sdk/typescript`
- `git diff --check`

GitHub reports the PR mergeable against `main`.

PR #1666, opened 2026-05-01
Move plugin discovery into plugins service

hughhan1:
## Summary

Moves plugin post-connect discovery execution out of `gestaltd/internal/discovery` and into `gestaltd/services/plugins/discovery`.

This keeps plugin discovery behavior under the plugin service-owned tree. Behavior is unchanged.

## Interface diff

```diff
- import "github.com/valon-technologies/gestalt/server/internal/discovery"
+ import "github.com/valon-technologies/gestalt/server/services/plugins/discovery"

  candidates, err := discovery.Run(ctx, cfg, client)
```

## Example

```go
import "github.com/valon-technologies/gestalt/server/services/plugins/discovery"

candidates, err := discovery.Run(ctx, provider.DiscoveryConfig(), client)
if err != nil {
    return err
}
```

## Testing

- `go test ./services/plugins/discovery`
- `go test ./internal/server -run 'Test.*Discovery|Test.*Connect|Test.*Manual|Test.*SessionCatalog|Test.*Composite|Test.*Restricted' -count=1`
- `go list ./services/plugins/discovery ./internal/server`
- `go list ./...`
- `git diff --check`


hughhan1:
Closing as obsolete: the plugin discovery move is already present on current main under gestaltd/services/plugins/discovery, so this older PR no longer has a unique patch after restacking.

PR #1694, opened 2026-05-01
Revert operator lifecycle package move

hughhan1:
## Summary

Reverts #1691 / commit `4e5fa7cab81b3e3cae2b700bd272672f6bdb9db2`.

- Move `operator` lifecycle code back from `gestaltd/internal/operator` to `gestaltd/services/operator`.
- Restore daemon imports to `github.com/valon-technologies/gestalt/server/services/operator`.
- Keep the #1690 `services/testutil` import fix intact.

## Rationale

`gestaltd/internal/operator` was idiomatic if we wanted the package to be daemon-private. But the name and role are still an operator service/lifecycle boundary, and `services/operator/internal` would not work for the current daemon call sites: Go internal import rules would restrict that package to code under `services/operator`, so `internal/daemon` could not import it directly without a public `services/operator` facade.

Given that, this revert keeps the operator package at the existing service-level boundary instead of moving it into top-level `internal` prematurely.

## Interface Changes

```diff
- import "github.com/valon-technologies/gestalt/server/internal/operator"
+ import "github.com/valon-technologies/gestalt/server/services/operator"
```

## Functional/Integration Checks

- `go test ./services/operator ./internal/daemon ./services/providerdrivers`
- `go mod tidy && git diff --exit-code go.mod go.sum`
- `go list ./...`
- `golangci-lint run ./services/operator ./internal/daemon ./services/providerdrivers`
- `git diff --check origin/main...HEAD`


hughhan1:
Closing per architecture decision: operator is daemon-owned/internal, so we should keep PR #1691's move to gestaltd/internal/operator rather than restoring gestaltd/services/operator.

PR #1712, opened 2026-05-02
Add authored SDK provider surfaces

hughhan1:
## Summary

- Move generated Go transport bindings under `internal/gen/v1` and keep `sdk/go/gen/v1` as a compatibility alias package for existing low-level Go users.
- Move Python generated stubs under `gestalt._gen` and TypeScript generated stubs under `src/internal/gen`, removing the public TS `gen` export.
- Add authored SDK surfaces for Go cache, S3, plugin runtime/logs, and IndexedDB provider adapters; add Python agent/workflow/authentication/authorization DTO helpers; add a TypeScript agent contract test helper export.
- Update custom-provider docs to steer provider authors toward authored SDK APIs instead of generated protobuf modules.

## Interface Changes

- New/changed interface: Go `CacheProvider`, `S3Provider`, `PluginRuntimeProvider`, `PluginRuntimeLogHost`, and `AuthoredIndexedDBProvider` DTO-style APIs.
- New/changed interface: Python public helpers such as `AgentSession`, `AgentInteraction`, workflow bound DTOs, authentication request DTOs, and authorization request DTOs.
- New/changed interface: TypeScript `@valon-technologies/gestalt/test/agent-contract` for downstream contract tests.
- Compatibility note: `github.com/valon-technologies/gestalt/sdk/go/gen/v1` remains as aliases for Go providers that still need protocol-level IndexedDB/workflow/authorization internals. New Go provider code should use authored SDK APIs where available.

Example usage:

```go
type Provider struct{ gestalt.BaseProvider }

func (p *Provider) HeadObject(ctx context.Context, ref gestalt.ObjectRef) (gestalt.ObjectMeta, error) {
    // provider-owned object lookup
}

var _ gestalt.S3Provider = (*Provider)(nil)
```

```ts
import { createAgentContractMessage } from "@valon-technologies/gestalt/test/agent-contract";
```

## Functional/Integration Tests

- Contract boundary covered: generated transports remain usable internally while authored SDK APIs serve cache/S3/runtime/IndexedDB and downstream Python/TS helpers.
- Commands run:
  - `git diff --check origin/main...HEAD`
  - `go test ./...` in `sdk/go`
  - `go test ./...` in `gestaltd`
  - `bun run typecheck` in `sdk/typescript`
  - `uv run ruff check .` in `sdk/python`
  - `uv run ty check --exclude 'gestalt/_gen/**' gestalt scripts tests` in `sdk/python`
  - `uv run python -m unittest discover -s tests` in `sdk/python`
[...]

PR #1719, opened 2026-05-02
fix docker image publish step

mzywang:
## Description

Fix the `gestaltd` Docker build after [#1712](https://github.com/valon-technologies/gestalt/pull/1712) introduced the root Go module replacement, which made the Docker dependency layer look for `/build/go.mod` before it had been copied.

The failure showed up in the [`Publish Docker Image (static)` job](https://github.com/valon-technologies/gestalt/actions/runs/25259881320/job/74065389590) and the [`Publish Docker Image (Alpine)` job](https://github.com/valon-technologies/gestalt/actions/runs/25259881320/job/74065389605). This change copies the root `go.mod`/`go.sum` before `go mod download` and adds CI Docker builds for both image targets so this is caught before release tagging.

### DIFF
--- a/.github/workflows/build-gestaltd.yml
+++ b/.github/workflows/build-gestaltd.yml
@@ -206,3 +206,32 @@ jobs:
 
       - name: Validate Kubernetes schemas (preparation)
         run: kubeconform -strict -summary /tmp/rendered-preparation.yaml
+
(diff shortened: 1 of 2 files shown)


mzywang:

- #1719 ◀

PR #1723, opened 2026-05-02
[gestaltd] remove Helm shell dependency and rename config path

mzywang:
## Description

Updates the gestaltd Helm chart preparation flow to run `gestaltd lock` and `gestaltd sync --locked` as separate init containers instead of using `/bin/sh -c`, which works with the static gestaltd image. Also renames the default container config path to `/etc/gestaltd/config.yaml` and updates the Docker and Helm docs to match.


mzywang:

- #1723 ◀

PR #1773, opened 2026-05-04
Select relevant default agent catalog tools

hughhan1:
## Summary
- refine default MCP catalog grants from a wildcard to prompt-ranked, operation-specific tool refs when the user turn has text
- keep the wildcard fallback when there is no query, no candidates, or refinement fails
- add issue/ticket search synonyms so prompts like "linear tickets" rank Linear issue tools

## Tests
- go test ./services/agents/agentmanager ./internal/server


hughhan1:
Closing per follow-up design direction. The prompt-ranking patch keeps semantic tool discovery in gestaltd core, and we want core to stay provider-agnostic while agent runtimes/providers own tool discovery.

PR #1793, opened 2026-05-05
Support platform OAuth refresh-token connections

hughhan1:
## Summary
- Add `auth.refreshToken` to Gestalt connection auth config, generated external-credential protobuf bindings, and runtime/declarative connection conversion.
- Allow top-level `mode: platform` OAuth2 connections to use `grantType: refresh_token` while keeping token resolution delegated to external credentials providers.
- Skip user OAuth handler construction for delegated refresh-token platform connections, and reject broker connection overrides that move an operation-bound call onto a different internal connection.
- Map invalid plugin-invoker requests to gRPC `InvalidArgument` instead of `Unknown`.
- Document deployer-owned refresh-token platform connections for internal plugin bindings.

## Test plan
- [x] `go test ./internal/config ./internal/bootstrap ./services/invocation ./services/externalcredentials ./services/plugininvoker` from `gestaltd/`
- [x] `git diff --check`
- [ ] `buf lint .` from `sdk/proto/` currently fails on existing repository-wide proto lint violations unrelated to this change

New interface: `connections.<id>.auth.refreshToken` is valid with `auth.type: oauth2`, `auth.grantType: refresh_token`, and `mode: platform`.

```yaml
connections:
  gmail-platform-mailbox:
    mode: platform
    auth:
      type: oauth2
      grantType: refresh_token
      tokenUrl: https://oauth2.googleapis.com/token
      clientId:
        secret:
          provider: secrets
          name: google-oauth-client-id
      clientSecret:
        secret:
          provider: secrets
          name: google-oauth-client-secret
      refreshToken:
        secret:
          provider: secrets
          name: gmail-platform-mailbox-refresh-token

plugins:
  gmail:
    connections:
      platform:
        ref: gmail-platform-mailbox
        exposure: internal
```

Contract coverage: deploy YAML -> connection runtime -> plugin invoker/broker -> external credentials transport. Follow-up PR: gestalt-providers will add refresh-token exchange support to the default external credentials provider and revise Gmail to consume the platform token through its connection contract.


hughhan1:
Fixed in the latest push: `validateTopLevelConnections` now rejects `auth.refreshToken` in the `client_credentials` branch, with a regression case in `TestValidateStructurePlatformOAuth2RefreshToken`. Targeted Go tests are passing again.

hughhan1:
Fixed the latest review finding in `5a3a3a413`: `buildConnectionHandler` now skips OAuth handler creation for both delegated OAuth grants (`client_credentials` and `refresh_token`), with `TestBuildConnectionAuthMapSkipsPlatformOAuthRefreshTokenHandler` covering the refresh-token path. I also added broker-side operation connection override enforcement so internal invocations match the public HTTP guard. Targeted Go tests are passing: `go test ./internal/config ./internal/bootstrap ./services/invocation ./services/externalcredentials ./services/plugininvoker` from `gestaltd/`.

hughhan1:
Pushed `167ba1c0b` after the final stack review. In addition to the refresh-token OAuth handler skip, broker invocation now rejects explicit overrides that would move an operation-bound call onto a different internal connection, even for fallback-bound declarative operations that otherwise allow user connection overrides. Plugin invoker also maps `ErrInvalidInvocation` to gRPC `InvalidArgument`.

Targeted verification remains green: `go test ./internal/config ./internal/bootstrap ./services/invocation ./services/externalcredentials ./services/plugininvoker` from `gestaltd/`, plus `git diff --check`.

PR #1805, opened 2026-05-06
Make agent session listing best effort across providers

hughhan1:
## Summary
- keep provider-filtered session listing strict, but allow unfiltered list calls to continue when one provider fails
- annotate provider list failures with the provider name so production logs identify the failing backend
- add regression coverage for best-effort unfiltered fan-out and strict provider-filtered failures

## Tests
- go test ./services/agents/agentmanager
- go test ./services/agents/...
- go test ./internal/server

### DIFF
--- a/gestaltd/services/agents/agentmanager/manager.go
+++ b/gestaltd/services/agents/agentmanager/manager.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
 	"maps"
 	"sort"
 	"strconv"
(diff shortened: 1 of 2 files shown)


hughhan1:
Closing this for now. The best-effort fan-out behavior would improve UI resilience, but it is a workaround for the current production failure. We should identify and fix the provider/runtime root cause first: summary session listing is still hitting a provider-side gRPC 4 MiB receive limit while reading roughly 8 MiB of session data.

hughhan1:
Closing while we investigate and fix the root cause instead of merging a best-effort workaround.

PR #1809, opened 2026-05-06
Make workflow config cleanup best effort

hughhan1:
## Summary
- Treat workflow schedule and event-trigger cleanup list failures as non-fatal during bootstrap.
- Log and skip cleanup for the affected workflow provider so desired config can still be reconciled and the service can start.
- Add regression coverage for both removed schedule cleanup and removed event-trigger cleanup when provider visibility/listing fails.

## Context
The toolshed deploy from valon-technologies/toolshed#1427 got past the schedule execution-ref outage handled by #1808, then repeatedly failed Cloud Run startup with:

```
bootstrap: bootstrap: list workflow event triggers for provider "local": rpc error: code = Internal desc = query temporal index shard 0: context canceled
```

Cleanup of removed config-owned workflow objects should not be an availability gate. Leaving stale config-owned schedules/triggers in place is safer than blocking startup; a later successful bootstrap can clean them up.

## Tests
- `go test ./internal/bootstrap`

### DIFF
--- a/gestaltd/internal/bootstrap/bootstrap_test.go
+++ b/gestaltd/internal/bootstrap/bootstrap_test.go
@@ -5261,6 +5261,78 @@ func TestBootstrapDeletesRemovedConfiguredWorkflowSchedules(t *testing.T) {
 	}
 }
 
+func TestBootstrapSkipsRemovedConfiguredWorkflowScheduleCleanupWhenListFails(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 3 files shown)


hughhan1:
Closing as superseded. #1811 already merged the bootstrap cleanup guard, and the root-cause fix is the Temporal workflow provider alpha.7 IndexedDB-backed state migration now deployed in toolshed.

PR #1810, opened 2026-05-06
Skip workflow config cleanup when indexes are unavailable

hughhan1:
## Summary
- Skip stale workflow schedule/event-trigger cleanup when the workflow provider cannot list indexed objects during bootstrap.
- Apply the existing-object fallback to event triggers as well as schedules when execution-ref refresh fails but the provider object still matches config.
- Add regression coverage for schedule and event-trigger cleanup list failures.

## Test plan
- [x] `go test ./internal/bootstrap`


hughhan1:
Closing as redundant: the cleanup startup fix landed on main via #1811, and this PR's remaining changes are not needed for the current Valon Tools deploy.

PR #1820, opened 2026-05-06
Break hosted workflow pool startup cycle

hughhan1:
## Summary
- stop creating a hosted workflow control provider during bootstrap for pooled hosted workflow providers
- use a local bootstrap control provider only while config-managed workflow schedules/triggers reconcile, then close it before workflow startup completes
- start long-running hosted workflow workers asynchronously and route workflow/control RPCs to ready runtime workers after the pool is up

## Why
Cloud Run cannot route public host-service relay traffic to a revision until the revision is ready. The old pooled workflow path configured a hosted control provider during bootstrap, so Temporal opened IndexedDB/workflow host services through the public relay before the new revision could receive that traffic. Requests hit the previous revision and failed with stale/unknown relay sessions.

This keeps steady-state workflow workers in the hosted runtime while avoiding the Cloud Run readiness cycle.

## Verification
- go test ./internal/bootstrap
- go test ./internal/server ./services/workflows

### DIFF
--- a/gestaltd/internal/bootstrap/bootstrap.go
+++ b/gestaltd/internal/bootstrap/bootstrap.go
@@ -1784,7 +1784,26 @@ func buildWorkflow(ctx context.Context, name string, entry *config.ProviderEntry
 	}
 	var provider coreworkflow.Provider
 	if entry.UsesHostedExecution() {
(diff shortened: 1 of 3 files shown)


hughhan1:
Superseded by #1821, which replaces this branch with the cleaner control-provider + runtime-worker abstraction and removes the hosted workflow control proxy.

PR #1835, opened 2026-05-07
Fix IndexedDB SDK docs mismatches

hughhan1:
## Summary
- clarify that the Go SDK has the high-level typed IndexedDB provider authoring surface
- remove stale TypeScript/Rust docs references to non-existent IndexedDB provider helpers
- update IndexedDB plugin-side Rust examples to match the current Rust SDK
- keep IndexedDB plugin SDK examples linear by avoiding deletion of the row before later queries/transactions

## Tests
- npm run build (docs)
- npm run typecheck (docs)
- git diff --check

### DIFF
--- a/docs/content/custom-providers/indexeddb.mdx
+++ b/docs/content/custom-providers/indexeddb.mdx
@@ -52,189 +52,87 @@ spec:
 
 ## Provider interface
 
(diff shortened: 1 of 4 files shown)


hughhan1:
Closing this docs-only PR. The narrower fix is not enough; the right follow-up is to add proper high-level SDK surfaces where provider code is currently forced down to generated gRPC stubs.

PR #1859, opened 2026-05-08
[gestaltd] defer stage chart reader grant

mzywang:
## Description

Remove the stage valon-tools Terraform service account from the default chart reader grants until the stage bootstrap creates that service account.


mzywang:

- #1859 ◀

PR #1897, opened 2026-05-09
Hide protobuf timestamp and empty helpers behind SDK boundaries

hughhan1:
## Summary
- replace Go external credential aliases with native structs using time.Time pointers while converting to generated protos only at transport boundaries
- add a provider lifecycle probe helper so providers do not construct Empty requests for readiness checks
- remove public Empty aliases/re-exports from Go, Python, TypeScript, and Rust SDK surfaces
- update protocol tests and SDK/provider docs to steer authors toward native inputs rather than protobuf WKT constructors

## Verification
- go test ./... (sdk/go)
- bun run check (sdk/typescript)
- uv run pytest (sdk/python)
- cargo test (sdk/rust)
- uv run sphinx-build -W -b html docs docs/_build/html (sdk/python)
- bun run docs:build (sdk/typescript)
- cargo doc --no-deps (sdk/rust)

### DIFF
--- a/docs/content/custom-providers/external-credentials.mdx
+++ b/docs/content/custom-providers/external-credentials.mdx
@@ -47,9 +47,9 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 
 	gestalt "github.com/valon-technologies/gestalt/sdk/go"
-	gproto "google.golang.org/protobuf/proto"
 )
 
 type Provider struct {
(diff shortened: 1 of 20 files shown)


hughhan1:
Pushed a fix for the Go CI fixture failure in gestaltd: the generated external-credentials provider fixture now uses the native SDK field names and *time.Time values introduced by this PR.\n\nFocused checks:\n- go test ./internal/operator -run '^$' -count=1\n- go test ./internal/daemon -run '^$' -count=1

PR #1901, opened 2026-05-10
Use native workflow builder inputs across SDKs

hughhan1:
## Summary
- make Go workflow builders accept native target, trigger, match, permission, run-as, output binding, agent message, and agent tool-ref inputs instead of requiring protocol messages
- add matching Rust native agent message/tool-ref workflow inputs and export the TS agent message/tool-ref input aliases
- keep protobuf conversion inside SDK builders and update Rust docs/tests plus Go tests

## Tests
- go test ./... (sdk/go)
- cargo test --test workflow_builders (sdk/rust)
- bun run typecheck (sdk/typescript)


hughhan1:
Update on scope: I pushed `ab0d5d6ab` to trim back the over-expanded agent message/tool-ref native mirror layer. The PR now keeps those on the existing agent protocol types and focuses native workflow builder inputs on the areas that reduce provider-side proto construction. Local checks passed: `go test ./...` in `sdk/go`, `cargo test --test workflow_builders` in `sdk/rust`, `bun run typecheck` in `sdk/typescript`, plus provider checks in `workflow/indexeddb` and `workflow/temporal`.

PR #1923, opened 2026-05-12
Split Go proto bindings by module

hughhan1:
## Summary
- remove the root Go module and move server generated protos/indexeddbcodec under `gestaltd/internal`
- generate a separate SDK-internal Go proto package under `sdk/go/internal`, with public `sdk/go/gen/v1` aliases for compatibility
- add CI/docs checks for server and SDK generation plus a dependency-closure guard to prevent one server test binary from linking both generated proto packages

## Verification
- `go test ./...` in `sdk/go`
- `go test ./...` in `gestaltd`
- `go test -count=1 ./internal/bootstrap ./internal/daemon` in `gestaltd`
- `bash ../.github/scripts/check-go-proto-split.sh .` in `gestaltd`
- regenerated server and SDK Go bindings with Buf and regenerated SDK aliases


hughhan1:
Follow-up pushed in `660b2a03545f70a42dc461723e1fadeab233557b`:

- Removed stale source-level `go_package` options that still referenced the deleted root Go module path.
- Regenerated Python SDK protobuf stubs so descriptors no longer carry the deleted root Go package option.
- Removed the stale duplicate `sdk/proto/gen` tree; the maintained Python generated package is `sdk/python/gestalt/_gen` and generation already uses a temp Buf output directory.
- Tightened `.github/scripts/check-go-proto-split.sh` to reject tracked references to the deleted root Go proto package path.
- Fixed `gestaltd/Dockerfile` so Docker builds no longer copy root `go.mod`/`go.sum`.

Additional verification run:

- `go test ./...` in `sdk/go`
- `go test ./...` in `gestaltd`
- `uv run python -m unittest discover -s tests` in `sdk/python`
- `bash .github/scripts/check-go-proto-split.sh gestaltd`
- `git diff --check`
- `DOCKER_BUILDKIT=1 docker build --file gestaltd/Dockerfile --target static --build-arg GESTALT_VERSION=ci --tag gestaltd:ci-static .`
- `DOCKER_BUILDKIT=1 docker build --file gestaltd/Dockerfile --target alpine --build-arg GESTALT_VERSION=ci --tag gestaltd:ci-alpine .`

hughhan1:
Follow-up for the `copyDir` duplication finding pushed in `970fbc7bc22908ad3798b0b560b4c7ba8d8ed3c4`:

- `CopyExampleProviderPlugin` now delegates to the same `copyDir` helper used by the SDK test-provider build path.
- That makes symlink handling consistent across both fixture-copy paths.

Verification:

- `go test ./internal/bootstrap ./internal/daemon ./services/plugins/providerpkg` in `gestaltd`
- `git diff --check`

On the PR size: the 36K additions are intentional generated-code churn from Option B, not hand-written scope creep. Current breakdown is:

- 34,557 additions in `sdk/go/internal/gen/v1` generated protobuf bindings
- 963 additions in public `sdk/go/gen/v1` alias/generator files
- 369 additions in the SDK-local `indexeddbcodec` copy
- 687 additions across the remaining hand-written code, CI, docs, Dockerfile, and tests

PR #1950, opened 2026-05-12
Publish gestaltd CI binary artifacts

hughhan1:
## Summary

- Adds a gated `gestaltd` CI artifact publisher behind the existing `Build Gestaltd` validation workflow.
- Adds a backfill interface that validates/builds/publishes an explicit full Gestalt commit SHA, with the packaging script loaded from the current workflow checkout so older commits can be backfilled.
- Adds Terraform-managed GCS storage, publisher identity, public-read artifact URLs, and repository-variable outputs for commit-addressed linux/amd64 binary artifacts.

## Interface Changes

- New/changed interface: `Build Gestaltd` accepts optional `workflow_dispatch` input `gestalt_ref` as a full 40-character commit SHA.
- New/changed interface: CI artifact publishing is enabled only when repo variable `GESTALTD_CI_ARTIFACT_PUBLISH_ENABLED=true`, so the initial landing path does not fail before infra/repo variables exist.
- New/changed interface: CI artifacts publish under `${GESTALTD_CI_ARTIFACT_BASE_URL}/<sha>/gestaltd-linux-x86_64.tar.gz`, with `.sha256` and `metadata.json` alongside it.
- Example usage: run the workflow manually with `gestalt_ref=dec35e9202bf63aa079db5893fd6ab4b6a17f521` to backfill that exact commit.

## Functional/Integration Tests

- Tests added or augmented: CI artifact packaging script verifies the deterministic version string and keeps the Alpine smoke enabled in GitHub Actions.
- Contract boundary covered: workflow ref resolution, older-commit backfill packaging, immutable artifact packaging metadata/checksum, Terraform storage/IAM configuration.
- Commands run: `actionlint .github/workflows/build-gestaltd.yml .github/workflows/deploy-gestaltd-artifacts.yml`; `shellcheck .github/scripts/build-gestaltd-ci-artifact.sh`; `terraform init -backend=false`; `terraform fmt -check`; `terraform validate`; backfill-style package/checksum for `dec35e9202bf63aa079db5893fd6ab4b6a17f521`; backfill-style package/checksum for `0ed12c33c673e5ecc900ac9275845cf0340aa4ad`.

## Test plan

- [x] `actionlint .github/workflows/build-gestaltd.yml .github/workflows/deploy-gestaltd-artifacts.yml`
- [x] `shellcheck .github/scripts/build-gestaltd-ci-artifact.sh`
- [x] `terraform init -backend=false`
- [x] `terraform fmt -check`
- [x] `terraform validate`
- [x] Backfill-style build/package/checksum for current Toolshed API pin with `GESTALTD_SKIP_ALPINE_SMOKE=1`
- [x] Backfill-style build/package/checksum for current Toolshed agent-runtime pin with `GESTALTD_SKIP_ALPINE_SMOKE=1`
- [ ] Alpine smoke via Docker locally; Docker daemon was not running on this machine, so this remains covered by the GitHub Actions publisher path

Follow-up: after this lands and the GCS bucket/repository variables are available, set `GESTALTD_CI_ARTIFACT_PUBLISH_ENABLED=true`, backfill the Valon Tools pins, and merge the Toolshed consumer PR.

### DIFF
--- a/.github/scripts/build-gestaltd-ci-artifact.sh
+++ b/.github/scripts/build-gestaltd-ci-artifact.sh
@@ -1,0 +1,104 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+source_dir="${GESTALTD_SOURCE_DIR:-$(git rev-parse --show-toplevel)}"
+source_dir="$(cd "${source_dir}" && pwd)"
+commit="${GESTALTD_COMMIT:?GESTALTD_COMMIT is required}"
+version="${GESTALTD_VERSION:?GESTALTD_VERSION is required}"
+output_dir="${GESTALTD_OUTPUT_DIR:-${source_dir}/dist/gestaltd-ci}"
+artifact_base_url="${GESTALTD_ARTIFACT_BASE_URL:-}"
+run_id="${GITHUB_RUN_ID:-local}"
+created_at="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+
+if [[ ! "${commit}" =~ ^[0-9a-f]{40}$ ]]; then
+  echo "GESTALTD_COMMIT must be a full lowercase 40-character commit SHA." >&2
+  exit 1
+fi
+
+if [[ "${output_dir}" != /* ]]; then
+  output_dir="${PWD}/${output_dir}"
+fi
+
+rm -rf "${output_dir}"
+mkdir -p "${output_dir}/build"
+
+(
+  cd "${source_dir}/gestaltd"
+  CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
+    -trimpath \
+    -ldflags "-X main.version=${version}" \
+    -o "${output_dir}/build/gestaltd" \
+    ./cmd/gestaltd
+)
+
+if [[ "$(uname -s)" == "Linux" && "$(uname -m)" == "x86_64" ]]; then
+  version_output="$("${output_dir}/build/gestaltd" version)"
+  if [[ "${version_output}" != "${version}" ]]; then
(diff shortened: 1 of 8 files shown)


hughhan1:
Follow-up from the second review pass: I updated the branch so manual `workflow_dispatch` backfills with an explicit `gestalt_ref` use a narrow build/package/upload path instead of running today's full lint/docs gates against older source commits. Main-branch publishes still remain gated behind the full existing validation jobs. I also added `roles/storage.admin` to the Terraform-managed deployer role set so the same root can create the new GCS bucket and bucket IAM bindings.

hughhan1:
Fixed the failed `Verify checked-in generated bindings` step from run 25761913391. The failure was caused by unpinned Buf remote plugins floating `buf.build/grpc/go` from `protoc-gen-go-grpc v1.6.1` to `v1.6.2`; the generated files only differed in the version header. I pinned the Go and gRPC remote plugins in `sdk/proto/buf.go.server.gen.yaml` to the versions that match the checked-in generated files and verified `buf generate --template buf.go.server.gen.yaml` now leaves `gestaltd/internal/gen/v1/*.pb.go` and `*_grpc.pb.go` unchanged locally.

hughhan1:
Follow-up on the generated-bindings failure: the failed PR run was testing the merge commit against current `main`, where the checked-in gRPC generated files are already at `protoc-gen-go-grpc v1.6.2`. I rebased the branch onto current `main` and pinned `sdk/proto/buf.go.server.gen.yaml` to `buf.build/grpc/go:v1.6.2`, then verified locally that `buf generate --template buf.go.server.gen.yaml` leaves the checked-in generated files unchanged.

PR #1991, opened 2026-05-13
Update demo video to use Gestalt CLI

hughhan1:
## Summary
- Re-record the Claude Code demo with Gestalt MCP disabled for the recorded Claude session.
- Force Claude to use an empty MCP config plus the Bash tool, so the demo invokes `gestalt plugin invoke ...` through the CLI.
- Inline the README demo as a video element instead of a plain link.

## Verification
- `npm run build:single` from `docs/`
- `ffprobe` confirmed the updated docs video is H.264/yuv420p at 1280x720 and 423 KB.
- Spot-checked the final video frame showing the expanded Bash transcript and HTTPBin POST result.


hughhan1:
Closing per request.

PR #2012, opened 2026-05-14
[gestaltd] include provider runtime libraries in alpine image

mzywang:
## Description
Adds Alpine libgcc and libstdc++ to the gestaltd image so source-backed provider binaries that depend on the musl C++ runtime, including Cursor, can start inside gestaltd.


mzywang:

- #2012 ◀

PR #2025, opened 2026-05-15
Add server-managed connection presets

hughhan1:
## Summary
- Add connection preset config for OAuth-backed connections, including server-only authorization params and expected metadata checks.
- Thread preset selection through OAuth start/state/callback so credentials are stored only after the provider metadata matches the selected preset.
- Expose safe preset status data in integration listing without leaking server-only preset policy fields.

## Test plan
- [x] `go test ./internal/config ./internal/server`
- [x] Independent implementation review for preset validation, OAuth state, and callback enforcement


hughhan1:
Closing after reworking the design down to the existing Slack multi-instance flow. Toolshed already leaves Slack user OAuth unpinned, so users can add separate Slack workspace instances without a new Gestalt core preset abstraction. The remaining implementation is provider-owned bot workspace token selection.

PR #2047, opened 2026-05-18
[gestaltd] support plugin operation sets

mzywang:
## Description

Adds provider manifest operation sets and lets deployments select an allowed operation set with default roles for authorization.


mzywang:
Closing per direction: we are reverting to the original valon-tools-only Rippling migration with inline allowedOperations.

PR #2049, opened 2026-05-19
Umbrella: durable stepped workflow contracts

hughhan1:
## Problem Statement

Closes #2090.

`steps` should be the generic workflow plan model for plugin and agent workflows. A workflow definition needs to express ordered actions, `when` guards, value binding, and per-step output state durably across manual runs, schedules, and event triggers.

## Solution

This PR is now an umbrella only. The implementation was split into three single-commit PRs so each layer can be reviewed and merged in order:

1. #2070 - make `BoundWorkflowTarget` steps-only and update normalization, validation, SDKs, CLI paths, and config parsing.
2. #2071 - add provider-owned durable workflow/action contracts, provider-owned execution-ref RPCs, and the `InvokeWorkflowAction` host callback boundary.
3. #2072 - normalize deployment terminology to definitions, remove the separate planning RPC, and keep execution refs/datastores provider-owned.

This branch intentionally contains only an empty commit on current `main`; all code review should happen in the downstream PRs above.

## Rollout Order

Merge #2070 first, then #2071, then #2072. Provider implementations can then target the final definition/action contract from #2072.


hughhan1:
Closing in favor of RFC tracking and the split implementation stack.

- **RFC:** #2090
- **Implementation PRs (merge in order):** #2070 → #2071 → #2072

This umbrella branch was intentionally an empty commit on `main`; all review happens in the phase PRs above.

PR #2060, opened 2026-05-19
Preserve agent workflow transport semantics

hughhan1:
## Summary

This PR contains the agent/workflow compatibility fixes that were separated from the IndexedDB factory work.

It preserves existing behavior around:

- explicit empty tool refs, so callers can intentionally disable default tool selection;
- per-turn timeout propagation across the manager service and SDK request builders;
- multi-step workflow agent targets across proto, config, HTTP schedule APIs, workflow runtime, and SDK helpers;
- workflow execution-reference permissions when steps introduce tool refs or output delivery.

## Split Notes

This PR is independent of the IndexedDB protocol and bootstrap PRs. It can be reviewed and merged on its own.

## Agent Manager Contract

The public agent turn surface continues to distinguish explicit empty tool refs from omitted tool refs, and preserves per-turn timeout propagation:

```diff
 type ManagerCreateTurnRequest struct {
     SessionID string
     Prompt string
     Messages []Message
     ToolRefs []ToolRef
+    ToolRefsSet bool
+    TimeoutSeconds int32
 }
```

Semantics:

- `ToolRefsSet == false && len(ToolRefs) == 0` means the provider/default tool source may be used.
- `ToolRefsSet == true && len(ToolRefs) == 0` means tools were explicitly set to empty and defaults must not be injected.
- `TimeoutSeconds` is forwarded to provider turn creation instead of being dropped by the manager layer.

The non-Go SDKs preserve the same wire behavior. For example, TypeScript request construction treats an explicit `toolRefs: []` as an intentional set:

```diff
 const request = {
   sessionId,
   prompt,
   messages,
   toolRefs,
+  toolRefsSet: input.toolRefs !== undefined,
+  timeoutSeconds: input.timeoutSeconds,
 }
```

Python follows the same distinction when serializing request dictionaries:

```diff
-tool_refs_set = data.get("tool_refs_set", False)
+tool_refs_set = data.get("tool_refs_set", False) or "tool_refs" in data
```

## Workflow Agent Step Contract

Workflow agent targets continue to support both a top-level single-turn target and a multi-step target:

```diff
 type WorkflowAgentTarget struct {
     ProviderName string
     Prompt string
     Messages []AgentMessage
     ToolRefs []AgentToolRef
+    Steps []WorkflowAgentStep
 }
```

The step shape remains part of the public workflow API:

```go
type WorkflowAgentStep struct {
    ID             string
    Prompt         string
    Messages       []AgentMessage
    ToolRefs       []AgentToolRef
    TimeoutSeconds int32
    OutputDelivery *WorkflowOutputDelivery
    When           *WorkflowStepCondition
}

type WorkflowStepCondition struct {
    StepID     string
    OutputPath string
    Equals     *structpb.Value
}
```

Validation keeps the previous behavior:

- A top-level agent target must provide prompt/messages unless it defines `Steps`.
- A stepped target may have empty top-level prompt/messages because each step supplies its own work.
- Step `when.equals` is considered present by field presence, so an explicit JSON/protobuf `null` is valid and round-trips.

The HTTP schedule/config and SDK shapes continue to expose `steps`. TypeScript keeps the public shape:

```ts
type WorkflowAgentStep = {
  id?: string
  prompt?: string
  messages?: AgentMessage[]
  toolRefs?: AgentToolRef[]
  timeoutSeconds?: number
  outputDelivery?: WorkflowOutputDelivery
  when?: {
    stepId?: string
    outputPath?: string
    equals?: unknown
  }
}
```

Python keeps helper support for explicit null conditions:

```python
WorkflowAgentStep(
    id="check",
    when=WorkflowStepCondition(
        step_id="previous",
        output_path="$.result",
        equals=None,
    ),
)
```

## Execution Permission Contract

When config-managed workflow steps add per-step tool refs or output delivery, the runtime permissions remain additive. The agent provider permission is preserved alongside step-specific plugin permissions:

```diff
 ExecutionReferencePermissions{
     Plugins: []PluginPermission{
+        {Plugin: target.Agent.ProviderName},
         {Plugin: stepTool.ProviderName, Operation: stepTool.OperationName},
     },
 }
```

A provider-level permission for the agent provider is preserved so `agentmanager.CreateSession` can still create the agent session before step-specific tools are evaluated.

## Test Plan

- `cd gestaltd && go test -count=1 ./services/agents/... ./internal/server ./internal/bootstrap -run 'Agent|WorkflowSchedule|ExecutionRef'`
- `cd sdk/go && go test -count=1 ./... -run 'AgentManager|IndexedDB'`
- `cd sdk/python && uv sync --frozen --group dev && uv run pytest -q tests/test_agent_transport.py tests/test_workflow_helpers.py`
- `cd sdk/typescript && bun install && bun test tests/agent-manager.transport.test.ts`
- `cd sdk/rust && cargo test agent_manager`

### DIFF
--- a/gestaltd/internal/bootstrap/agent_runtime_test.go
+++ b/gestaltd/internal/bootstrap/agent_runtime_test.go
@@ -1405,7 +1405,7 @@ func TestAgentRuntimeConfigKeepsHostedAgentServingWhenProactiveReplacementStartF
 			ExpiresAt: &expiresAt,
 		}
 		if index == 1 {
-			recommendedDrainAt := startedAt.Add(3 * time.Second)
+			recommendedDrainAt := startedAt.Add(8 * time.Second)
 			lifecycle.RecommendedDrainAt = &recommendedDrainAt
 		}
 		return lifecycle
@@ -1468,7 +1468,7 @@ func TestAgentRuntimeConfigKeepsHostedAgentServingWhenProactiveReplacementStartF
 		t.Fatalf("ready backends = %d, want 1", len(backends))
 	}
 	first := backends[0]
-	waitForAgentRuntimeCondition(t, 2*time.Second, func() bool {
+	waitForAgentRuntimeCondition(t, 6*time.Second, func() bool {
 		pool.mu.Lock()
 		defer pool.mu.Unlock()
 		return len(runtimeProvider.startSessionRequests()) >= 2 && !first.replacing
--- a/gestaltd/internal/bootstrap/bootstrap_test.go
+++ b/gestaltd/internal/bootstrap/bootstrap_test.go
@@ -6489,9 +6489,73 @@ func TestBootstrapConfigManagedAgentTargetsPreserveWorkflowSystemToolRefs(t *tes
 		if err != nil {
 			t.Fatalf("Get execution ref %q: %v", executionRef, err)
 		}
-		if len(ref.Permissions) != 1 || ref.Permissions[0].Plugin != "roadmap" || len(ref.Permissions[0].Operations) != 1 || ref.Permissions[0].Operations[0] != "sync" {
-			t.Fatalf("permissions = %#v", ref.Permissions)
+		wantPermissions := []core.AccessPermission{
+			{Plugin: "managed"},
+			{Plugin: "roadmap", Operations: []string{"sync"}},
 		}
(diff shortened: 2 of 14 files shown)


hughhan1:
Closing stale PR; local and remote branches removed after 2+ days of inactivity.

PR #2096, opened 2026-05-21
[codex] Document local development features

hughhan1:
## Summary

- Document recently shipped local development paths for scoped serve, config watch, remote attach, local source lock behavior, and parallel local source preparation.
- Add manifest reference coverage for prepare-only source builds and `run.commandPrefix`.
- Correct app and CLI reference wording around `spec.assetRoot` and `--artifacts-dir`.

## Validation

- `npm run build:single`
- `npm run typecheck`
- Started a local docs server from the built output and verified edited pages return `200 OK`.

### DIFF
--- a/docs/Dockerfile
+++ b/docs/Dockerfile
@@ -84,6 +84,10 @@ server {
         try_files $uri =404;
     }
 
+    location ~ ^/(dev|latest|versions/v[^/]+)/_pagefind/(.*)$ {
+        try_files /_pagefind/$2 =404;
+    }
+
     location /dev/ {
         try_files $uri $uri.html $uri/ =404;
     }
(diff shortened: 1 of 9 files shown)


hughhan1:
Update: added a prod search fix in `16de73d09`.

Root cause: versioned docs pages request Pagefind from paths such as `/latest/_pagefind/pagefind.js`, while the built Pagefind bundle is emitted at `/_pagefind/pagefind.js`. Production currently returns `404` for the versioned URL and `200` for the unversioned bundle.

Fix: nginx now maps `/dev/_pagefind/*`, `/latest/_pagefind/*`, and `/versions/v*/_pagefind/*` to the generated global `/_pagefind/*` assets. The static registry test host mirrors that behavior.

Validation:
- `node --check docs/scripts/check-nginx-routing.mjs`
- `node --check docs/scripts/check-registry-static.mjs`
- `npm run build:versioned:test`
- `npm run test:registry`
- `npm run test:nginx-routing`
- `npm run typecheck`

PR #2087, opened 2026-05-20
[authorization] add composer and reconciler

mzywang:
## Description

Composes configured and dynamic authorization fragments into the provider-backed model, reconciles synthesized relationships, and backfills legacy provider dynamic tuples into the IndexedDB fragment source before treating that source as authoritative.

PR #2086, opened 2026-05-20
[authorization] introduce dynamic fragment source store APIs

mzywang:
## Description

Adds the first Gestalt-owned dynamic authorization source layer. This PR creates the `authz_dynamic_fragments` IndexedDB store, adds versioned fragment read/write helpers, backfills existing `plugin_dynamic` and `admin_dynamic` provider relationships into source fragments, and updates existing plugin/admin member compatibility flows so future writes populate the source store while preserving current provider-backed behavior.

New admin APIs added:

- `GET /admin/api/v1/authorization/fragments` - list dynamic authorization fragments.
- `GET /admin/api/v1/authorization/fragments/{fragmentID}` - fetch one dynamic authorization fragment.
- `PUT /admin/api/v1/authorization/fragments/{fragmentID}` - create or replace one dynamic authorization fragment with optional `expectedVersion` optimistic concurrency.
- `DELETE /admin/api/v1/authorization/fragments/{fragmentID}` - delete one dynamic authorization fragment.

Compatibility behavior updated:

- `PUT /admin/api/v1/authorization/plugins/{plugin}/members` now also upserts the plugin-owned dynamic source fragment.
- `DELETE /admin/api/v1/authorization/plugins/{plugin}/members/{subjectID}` now also removes the matching plugin-owned source relationship.
- `PUT /admin/api/v1/authorization/admins/members` now also upserts the global dynamic source fragment.
- `DELETE /admin/api/v1/authorization/admins/members/{subjectID}` now also removes the matching global source relationship.

This PR does not move authorization evaluation or reconciliation to the new source store yet; provider relationship writes remain in place for access-decision parity until the composer/reconciler PR.

PR #2100, opened 2026-05-21
[authorization] fix service account authz provider state

mzywang:
## Description

Allows default-allow policies to grant viewer access to authenticated users and service-account subjects only, and makes dynamic plugin/admin member deletion remove persisted fragment state so reloads do not resurrect removed members.

### DIFF
--- a/gestaltd/internal/coredata/authz_dynamic_fragments.go
+++ b/gestaltd/internal/coredata/authz_dynamic_fragments.go
@@ -266,17 +266,7 @@ func (s *AuthorizationDynamicFragmentService) ReplaceSubjectResourceRelationship
 		if err := validateAuthorizationFragmentRelationship(relationship); err != nil {
 			return err
 		}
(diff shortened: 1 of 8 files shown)


mzywang:

- #2084
    - #2086
        - #2087
            - #2100 ◀

hughhan1 (approved):

mzywang:
Addressed the structural feedback in 857a466b:

- extracted shared subject+resource relationship filtering and removed the pre-scan path
- centralized empty-fragment GC for both exact relationship deletes and subject/resource deletes
- narrowed and documented default-allow to users plus service accounts, with focused authorization unit coverage
- added a shared provider dynamic membership delete helper and warning for provider/fragment mismatch
- moved the delete/reload regression out of server_test.go and replaced the service-account default-allow integration coverage with a targeted authorization unit test

Local checks run:
- `go test ./...` from `gestaltd`
- `.github/scripts/check-go-proto-split.sh gestaltd`

`golangci-lint` is not installed locally, so I am relying on the PR CI lint job for that check.

Issue #2090, opened 2026-05-20
RFC: Durable stepped workflow definitions and provider action contracts


hughhan1:
## Problem Statement

Gestalt workflow targets are either a **single app call** or an **agent turn with optional agent-scoped substeps**. That split makes common multi-step workflows awkward to express. For example, [`OutputDelivery`](https://github.com/valon-technologies/gestalt/blob/1dee40fb29b5a91de53820b8db271c1a82197349/gestaltd/core/workflow/workflow.go#L84) was a workaround to emulate follow-on workflow behavior, but it is hard to reason about and complex.

A typical incident digest is app → agent → app in one definition:

```yaml
steps:
  - id: diagnose
    app:
      name: datadog
      operation: incidents.search

  - id: summarize
    agent:
      provider: openai
      prompt: "Summarize ${{ steps.diagnose.outputs.incidents }}."

  - id: notify
    app:
      name: slack
      operation: replies.create
      input:
        text: ${{ steps.summarize.outputs }}
```

This is not representable as a single workflow target.

Gestalt needs to let authors declare these multi-step flows in one workflow definition and execute them durably across manual runs, schedules, and event triggers.

There is also a separate architecture problem: workflow control is more complex than necessary. App/SDK calls currently enter [`WorkflowManagerHost`](https://github.com/valon-technologies/gestalt/blob/f8a27517e48d479d17c81cd2361c3a95867748f2/sdk/proto/v1/workflow.proto#L617), pass through [`ManagerServer`](https://github.com/valon-technologies/gestalt/blob/f8a27517e48d479d17c81cd2361c3a95867748f2/gestaltd/services/workflows/manager_server.go#L29) and [`workflowmanager.Service`](https://github.com/valon-technologies/gestalt/blob/f8a27517e48d479d17c81cd2361c3a95867748f2/gestaltd/services/workflows/workflowmanager/manager.go#L71), then delegate to [`WorkflowProvider`](https://github.com/valon-technologies/gestalt/blob/f8a27517e48d479d17c81cd2361c3a95867748f2/sdk/proto/v1/workflow.proto#L584), while side-effect execution is modeled as a separate workflow-specific callback surface. The target model and service model should both collapse to fewer primitives: the trusted `WorkflowProvider` owns durable orchestration and uses normal Gestalt SDK/API calls to execute step side effects.

### Requirements

- **Ordered multi-step workflows.** A single workflow definition can declare a sequence of actions. Steps run in declaration order, one at a time.
[...]

PR #2113, opened 2026-05-22
Remove workflow output delivery

hughhan1:
## Problem

Closes #2090.

Workflow step `outputDelivery` is a hidden second side effect attached to a producing step. That conflicts with the steps model we want: each side effect should be an explicit workflow step with its own id, inputs, guard, timeout, permissions, and failure state.

`deliverResultToCaller` also preserved that special case through inherited agent grants, which made agent-authored workflows harder to reason about and kept delivery outside the normal step surface.

## Solution

Remove `WorkflowStep.output_delivery`, `WorkflowStepDelivery`, and the hidden `deliverResultToCaller` path.

Delivery is now modeled as an ordinary plugin step that references a previous step with `stepOutput`:

```yaml
steps:
  - id: summarize
    agent:
      provider: openai
      model: gpt-5.5
      prompt:
        template: "Summarize ${inputs.incidents}."

  - id: reply
    plugin:
      name: slack
      operation: replies.create
      input:
        object:
          text:
            stepOutput:
              stepId: summarize
              path: text
```

This keeps side effects one-per-step and lets normal step ordering, permissions, and output references handle delivery.

## Interface Changes

```diff
 message WorkflowStep {
+  reserved 5;
+  reserved "output_delivery";
   string id = 1;
   map<string, WorkflowValue> inputs = 2;
   WorkflowStepWhen when = 3;
   int32 timeout_seconds = 4;
-  WorkflowStepDelivery output_delivery = 5;
   google.protobuf.Struct metadata = 6;
 
   oneof action {
     WorkflowStepPluginCall plugin = 10;
     WorkflowStepAgentTurn agent = 11;
   }
 }
-
-message WorkflowStepDelivery {
-  WorkflowStepPluginCall plugin = 1;
-}
```

SDK helper types and REST/config conversion follow the same deletion.

## Compatibility Note

This intentionally removes the special-cased delivery API. Existing workflows should use an explicit plugin step for Slack replies or any other delivery action.

`WorkflowHost.InvokeOperation` stays for this slice because the current Temporal and IndexedDB providers still use it for whole-target execution. Deleting that callback path belongs with provider-owned durable step interpretation.

## Validation

- `cd gestaltd && go test ./core/workflow ./internal/config ./internal/bootstrap ./internal/server ./services/agents/agentgrant ./services/agents/agentmanager ./services/workflows ./services/workflows/workflowmanager`
- `cd sdk/go && go test ./...`
- `cd sdk/python && uv run --frozen python -m unittest tests.test_workflow_helpers`
- `cd sdk/python && uv run --frozen sphinx-build -W -b html -d docs/_build/doctrees docs docs/_build/html`
- `cd sdk/typescript && bun run check`
- `cd sdk/rust && cargo test`
- `cd gestalt/cli && cargo test --test workflows_cli`
- `git diff --check`


hughhan1:
Status note: all normal CI checks are passing on this PR after rerunning the transient TypeScript SDK flake. The only failing check is `Gestalt Review`, whose check-run output is `Review failed` / `Gestalt review failed before it could complete.` with no annotations. I also tried GitHub check-run rerequest for the failed app check and GitHub returned 404, so this appears to need app/admin intervention rather than a code change.

hughhan1:
Closing as superseded by the rebased workflow stack: #2124 carries the coherent step-target/no-output-delivery end state, with follow-up cleanup in #2126 and #2114.

PR #2132, opened 2026-05-23
Drop legacy workflow proto reserved names

hughhan1:
## Summary
- Remove retired workflow field-name reservations from `workflow.proto` while keeping numeric reservations for wire safety.
- Regenerate checked-in workflow bindings for server Go, Go SDK, Python SDK, and TypeScript SDK.
- Remove remaining legacy workflow wording from touched tests while preserving caller-claim coverage.

## Validation
- `git diff --check`
- tracked-source search for retired workflow strings returned no matches
- `go test ./services/appinvoker ./internal/daemon` from `gestaltd`
- `go test ./...` from `sdk/go`
- `bun install --frozen-lockfile && bun run check` from `sdk/typescript`
- `uv run --group dev pytest tests/test_workflow_helpers.py tests/test_workflow_transport.py tests/test_runtime.py` from `sdk/python`

Note: I also tried `buf lint` from `sdk/proto`; it still fails on the repo's existing lint baseline (package path/service/RPC naming), not on this change.


hughhan1:
CI status note: all ordinary checks are green on `2a71f0b71712e12e1a49baff396ab452c419b5b4`. The only failure is `Gestalt Review`:

- https://github.com/valon-technologies/gestalt/runs/77553059422
- The check output says: `Gestalt review failed before it could complete.`
- It produced zero annotations, so I do not see a code-level finding to address in this PR.

hughhan1:
Follow-up after Toolshed audit: pushed `c7072110a` to fix TypeScript workflow step action narrowing for loose TypeScript consumers. This was exposed when `valon-tools/authentication/local` consumed `@valon-technologies/gestalt@0.0.1-alpha.23` under `strict: false`.

Validation added/run:
- `bun run check` in `sdk/typescript` now runs both the normal strict typecheck and `typecheck:loose`
- Refreshed CI is green for all ordinary checks on `c7072110af164df608cd5d12e94178c89053f695`
- The only failing check remains `Gestalt Review`, which again failed before completing: https://github.com/valon-technologies/gestalt/runs/77554261326

PR #2133, opened 2026-05-24
Migrate authz dynamic fragments to app schema

hughhan1:
## Summary
- migrate legacy `authz_dynamic_fragments` stores from the plugin schema to the app schema during coredata startup
- rewrite plugin-owned fragment records to app-owned records, including `plugin_dynamic` relationship/resource type references
- keep unknown schema mismatches failing instead of silently changing them

## Testing
- `go test ./internal/coredata`
- `go test ./internal/coredata ./internal/server -run 'TestNewMigratesLegacyAuthorizationDynamicFragmentsPluginSchema|AuthorizationDynamic|AdminAPI_DeletePluginMember|ProviderBackedReload'`\n- `go test ./services/authorization`\n\nContext: Valon Tools post-merge deploy is currently failing because production has the old `authz_dynamic_fragments` plugin schema and the alpha.24 server expects the app schema.

### DIFF
--- a/gestaltd/internal/coredata/authz_dynamic_fragments_migration.go
+++ b/gestaltd/internal/coredata/authz_dynamic_fragments_migration.go
@@ -1,0 +1,184 @@
+package coredata
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/valon-technologies/gestalt/server/core/indexeddb"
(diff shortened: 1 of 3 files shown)


hughhan1:
Closing per operator direction. We are running the targeted Valon Tools Cloud SQL migration directly instead of checking this PR in.

PR #2170, opened 2026-05-26
Test workflow signal rendering for agent turns

hughhan1:
## Summary
- Add an executor-level regression test for workflow agent signal payload rendering.
- Verify `${signalPayload.user_prompt}` is rendered into both configured agent messages and the prompt-derived user message before `AgentCreateTurn` is sent.
- Guard the workflow-to-agent boundary that Slack and GitHub provider-generated prompts depend on.

## Test plan
- [x] Run `cd /Users/hugh/src/gestalt-workflow-signal-guardrail/sdk/go && go test ./workflow`.
- [x] Run independent implementation review; no findings.

### DIFF
--- a/sdk/go/workflow/workflow_test.go
+++ b/sdk/go/workflow/workflow_test.go
@@ -87,6 +87,59 @@ func TestExecutorInvokesAppStep(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


hughhan1:
Closing this as unnecessary for the provider root fix. The actual change is in gestalt-providers: provider-generated workflow agent prompts now render signal payload templates directly, and no SDK behavior change is required.

PR #2239, opened 2026-05-28
Use workflow owner identities

hughhan1:
Summary:
- Replace config-managed workflow `runAs` with an explicit `owner` subject on schedules and event triggers.
- Use the workflow owner as the execution identity when minting workflow invocation tokens and upserting config-managed workflow objects.
- Keep workflow agent tool refs from carrying per-tool runAs delegation, with improved target path errors.

Test plan:
- `go test ./gestaltd/core ./gestaltd/internal/workflowwire`
- `go test ./gestaltd/internal/config ./gestaltd/services/workflows/workflowmanager`
- Focused `gestaltd/internal/bootstrap` owner/tool-ref tests
- Focused `gestaltd/internal/operator` scoped-load tests

Note:
- Full `go test ./gestaltd/internal/bootstrap` still has unrelated hosted-agent git clone failures in this local environment.

### DIFF
--- a/gestaltd/core/subjects.go
+++ b/gestaltd/core/subjects.go
@@ -1,6 +1,11 @@
 package core
 
-import "strings"
+import (
+	"fmt"
+	"strings"
+)
+
+const SubjectKindServiceAccount = "service_account"
 
 // Actor identifies who performed an action without credential-scoping fields.
 type Actor struct {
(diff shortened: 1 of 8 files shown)


hughhan1:
Dropping this PR: static workflow owner config is unnecessary for this path. Dynamic workflow triggers already carry the creator identity, and the execution behavior is moving to trigger-owner semantics in gestalt-providers instead.

PR #2249, opened 2026-06-01
Thread --platform through committed lock phase 1

hughhan1:
## Summary

`gestaltd lock --platform` previously ran in two phases where only phase 2 respected the flag. Phase 1 still materialized packages and embedded lock `manifest` metadata using `runtime.GOOS`/`GOARCH`, so a macOS developer locking for `linux/amd64` could commit host-specific archives and drift from CI.

Committed-lock preparation now resolves a materialization platform from the first parsed `--platform` value (or the host when the flag is omitted) and threads it through metadata installs, git snapshot locking, `applyLocked*`, and `materializeLocked*`. Phase 2 continues to hash any additional platforms in the same invocation. Sync and runtime execution still materialize on the host.

Lock and check error messages now suggest the same `--platform` flag via `formatLockFlags`, so remediation matches the command Toolshed and deploy docs use.

## Interfaces

```bash
gestaltd lock --config gestalt.yaml --config prod/config.yaml --platform linux/amd64
gestaltd lock --check --config gestalt.yaml --platform linux/amd64
```

```go
// gestaltd/internal/operator/lock_platform.go
func lockMaterializationPlatform(platforms []struct{ GOOS, GOARCH string }) string
func resolveMaterializationPlatform(platform string) string
func pathsMaterializationPlatform(paths lifecyclePaths) string
```

When multiple `--platform` values are passed, phase 1 uses the first; phase 2 still hashes every listed platform.

## Runtime

`LockAtPathsWithPlatforms` and `CheckLockAtPathsWithStatePaths` call `prepareCommittedLockAtPathsInScratch` with the resolved platform, store it on `lifecyclePaths.lockMaterializationPlatform`, and pass it into `installMetadataSourcePackage` so phase 1 downloads and installs the requested archive before `attachStaticValidationMetadata` copies `ResolvedManifest` into the lock entry.

Sync-only call sites continue to pass an empty platform string, which falls back to `CurrentPlatformString()` via `resolveMaterializationPlatform`.


hughhan1:
Closing as superseded by #2292, which makes committed lock manifests platform-neutral instead of persisting a materialization platform and threading `--platform` through phase 1 lock preparation.

PR #2280, opened 2026-06-03
Retry stale GCS release metadata with cache busting

hughhan1:
## Summary
- Retries public GCS `provider-release.yaml` fetches with a cache-busting query only when the first response is stale metadata missing required `staticValidation`.
- Keeps strict `staticValidation` validation for non-GCS metadata and other decode errors.

## Test plan
- [x] `TMPDIR=/tmp go test ./gestaltd/internal/staticvalidation ./gestaltd/internal/operator ./gestaltd/internal/daemon`

### DIFF
--- a/gestaltd/internal/operator/release_metadata.go
+++ b/gestaltd/internal/operator/release_metadata.go
@@ -238,9 +238,57 @@ func fetchProviderReleaseMetadata(ctx context.Context, client *http.Client, meta
 	}
 	metadata, err := decodeProviderReleaseMetadata(data)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


hughhan1:
Closing as superseded.

Main now handles this in two ways without the retry hack:

1. **#2312** — self-contained `provider-release.yaml` with embedded `staticValidation` (no sidecar indirection).
2. **Git snapshot URLs** — `git_source.go` always appends a `sourceRef` query param when resolving GCS snapshot metadata URLs, which busts CDN cache on fetch.

Deploy snapshots were also republished at the existing lock refs with gestaltd `328dad983`, so stale `#2303`-era metadata should no longer be served.

PR #2283, opened 2026-06-03
Honor agent turn create timeouts

hughhan1:
## Summary

Agent turn creation now uses the caller's explicit turn timeout when deciding how long the provider RPC may run, instead of always falling back to the generic provider RPC timeout for app-originated turns.

Workflow-owned agent turns keep the existing marked workflow deadline behavior. Unmarked turn creation without an explicit timeout still uses the normal provider RPC timeout, so short provider calls remain bounded by the existing default.

This lets direct app calls and workflow agent steps share the same timeout contract through `CreateAgentProviderTurnRequest.timeout_seconds`.

## Interfaces

```proto
message CreateAgentProviderTurnRequest {
  int32 timeout_seconds = 18;
}
```

```go
func ProviderAgentTurnCreateContext(parent context.Context, timeoutSeconds int32) (context.Context, context.CancelFunc)
```

## Runtime

`remoteAgent.CreateTurn` now builds its provider call context with `ProviderAgentTurnCreateContext`, passing through the request timeout. The helper preserves marked workflow deadlines first, applies a positive request timeout for direct turn creation, and otherwise falls back to `ProviderCallContext`.

The existing unmarked provider timeout test now also verifies the request-timeout path, so the behavior is covered without adding a separate test case.

## Test plan

- [x] `go test ./gestaltd/services/runtimehost ./gestaltd/services/agents`

### DIFF
--- a/gestaltd/services/agents/client.go
+++ b/gestaltd/services/agents/client.go
@@ -145,7 +145,7 @@ func (r *remoteAgent) UpdateSession(ctx context.Context, req *proto.UpdateAgentP
 }
 
 func (r *remoteAgent) CreateTurn(ctx context.Context, req *proto.CreateAgentProviderTurnRequest) (*coreagent.Turn, error) {
-	ctx, cancel := runtimehost.ProviderWorkflowAgentCallContext(ctx)
+	ctx, cancel := runtimehost.ProviderAgentTurnCreateContext(ctx, req.GetTimeoutSeconds())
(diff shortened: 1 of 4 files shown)


hughhan1:
Closing this stale branch. The replacement keeps the existing timeout_seconds field and synchronous provider behavior, but reapplies the CreateTurn RPC deadline change from a clean origin/main branch with focused tests.

PR #2284, opened 2026-06-03
Honor agent turn create timeouts

hughhan1:
## Summary

Agent turn creation now uses the existing `CreateAgentProviderTurnRequest.timeout_seconds` budget when deciding how long a direct provider `CreateTurn` RPC may run.

Workflow-marked agent calls keep their existing deadline behavior: marked calls with a parent deadline preserve that parent deadline, and marked calls without one fall back to the generic provider RPC timeout.

This replaces the closed `hugh/agent-turn-create-timeout` branch from a clean `origin/main` base, keeping the already-published timeout interface and tightening the helper and workflow coverage around it.

## Interfaces

```proto
message CreateAgentProviderTurnRequest {
  int32 timeout_seconds = 18;
}
```

```go
func ProviderAgentTurnCreateContext(parent context.Context, timeoutSeconds int32) (context.Context, context.CancelFunc)
```

## Runtime

`remoteAgent.CreateTurn` now builds its provider call context with `ProviderAgentTurnCreateContext`, passing the request timeout through to the runtime helper.

The helper applies positive request timeouts only for unmarked direct turn creation. Workflow-marked calls continue to use the existing workflow-agent deadline semantics so workflow execution remains bounded by the workflow context rather than by a second request-derived deadline.

The existing Go workflow executor test now also asserts that workflow step `timeoutSeconds` is passed into `AgentCreateTurn`, which covers the SDK step invocation path without adding app-specific fixtures.

## Test plan

- [x] `go test ./gestaltd/services/runtimehost ./gestaltd/services/agents ./sdk/go/workflow`

### DIFF
--- a/gestaltd/services/agents/client.go
+++ b/gestaltd/services/agents/client.go
@@ -145,7 +145,7 @@ func (r *remoteAgent) UpdateSession(ctx context.Context, req *proto.UpdateAgentP
 }
 
 func (r *remoteAgent) CreateTurn(ctx context.Context, req *proto.CreateAgentProviderTurnRequest) (*coreagent.Turn, error) {
-	ctx, cancel := runtimehost.ProviderWorkflowAgentCallContext(ctx)
+	ctx, cancel := runtimehost.ProviderAgentTurnCreateContext(ctx, req.GetTimeoutSeconds())
 	defer cancel()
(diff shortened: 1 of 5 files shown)


hughhan1:
Superseded by the async CreateTurn contract work; this synchronous timeout direction is intentionally not being pursued.

PR #2297, opened 2026-06-04
[auth] hydrate subject email for canonical users

mzywang:
## Description
Hydrates broker user principals from the user store when a request already has a canonical user subject but lacks identity email, so app provider request contexts include `subject.email` for user API/session principals.

Tested with `go test ./gestaltd/services/invocation`.


mzywang:
Closing per author request; keeping #2309 open.

PR #2306, opened 2026-06-04
Prefer workflow runAs for host-service calls

hughhan1:
## Summary
- Use workflow provider/run metadata instead of forwarding short-lived invocation tokens for workflow-originated TypeScript SDK host-service calls.
- Preserve explicit string-token behavior for non-request callers.
- Unblock long-running workflow steps from losing agent/app access after the root invocation token TTL.

## Test plan
- [x] bun run typecheck in sdk/typescript


hughhan1:
Closing this tactical patch. The right fix is to remove invocationToken as the cross-service authorization concept rather than special-case workflow calls in the TypeScript SDK.

PR #2353, opened 2026-06-08
[credentials] allow service account external credential provisioning

mzywang:
## Description

Adds serviceAccountId support to OAuth and manual connection flows so external credentials can be stored under a managed service account subject.


mzywang:

- #2353 ◀

PR #2379, opened 2026-06-10
Rewire the handwritten cache client through the generated SDK

### LINKED ISSUE #2363: RFC: Generated, idiomatic, language-native SDKs
## Problem Statement

Gestalt's provider SDKs must be ergonomic to use, consistent across languages, and stay synchronized with the proto contract as it evolves.

Per-language wire-stub generators (such as [`protobuf-es`](https://github.com/bufbuild/protobuf-es), [`prost`](https://github.com/tokio-rs/prost), [`betterproto`](https://github.com/danielgtaylor/python-betterproto), and [`protoc-gen-go`](https://github.com/protocolbuffers/protobuf-go), coordinated by [`buf`](https://buf.build/)) keep SDKs in sync with the proto but produce wire-shaped types and gRPC stubs rather than ergonomic SDK surfaces, and each makes independent semantic choices for oneofs, enums, and well-known types that diverge across languages.

Handwritten domain SDKs avoid this but make four language surfaces drift on field naming, enum semantics, oneof behavior, timestamp conversion, JSON object handling, and conversion logic.

## Requirements

- **Single source of truth.** Changes to the proto produce SDK changes without manual edits.
- **Idiomatic SDKs.** SDKs expose idiomatic types and construction patterns, not wire-oriented proto message structs.
- **Proto3 fidelity.** SDKs represent proto3 semantics ([presence](https://protobuf.dev/programming-guides/field_presence/), oneofs, enums, well-known types) across all languages.
- **Transport flexibility.** SDKs are independent of transport so that gRPC and REST can coexist without distorting SDK shape.
- **Deterministic output.** Generation is byte-stable and drift is enforced by CI.

### Non-Goals

- Replacing wire stubs. The wire layer remains Buf/protoc-generated.
- Preserving backwards compatibility. Breaking changes are allowed; no aliases or fallbacks.

## Proposed Solution

Introduce `sdkgen`, a descriptor-driven generator for provider SDKs.

The protobuf descriptor remains the source of truth. Existing Buf/protoc-generated wire stubs remain the transport/wire layer. `sdkgen` reads a compiled proto schema, discovers supported provider services by convention, builds a small normalized model, and emits native SDK types, conversions, clients, and provider/server scaffolding for TypeScript, Python, Go, and Rust. The compiled proto schema is implemented as a Buf descriptor image generated during local generation and CI.

If a provider uses a proto construct the generator does not support yet, generation fails with a clear error.

`sdkgen` is a single binary rather than a set of per-language protoc plugins. The normalized model is built once and shared by every emitter, so emitters cannot diverge on semantics, and the generator sees its complete output file set across all four SDKs, which the file-ownership rules below depend on.

### Architecture

```mermaid
flowchart TD
  proto[.proto sources] -->|buf build| image[descriptor image]
  image -->|buf templates| wire[wire stubs]
  subgraph sdkgen
    validator --> model[normalized model]
    model --> ts[TS emitter]
    model --> py[Python emitter]
    model --> go[Go emitter]
    model --> rs[Rust emitter]
  end
  image --> validator
  ts --> out[SDKs]
  py --> out
  go --> out
  rs --> out
  wire --> out
  out --> ci[CI drift check]
```

The descriptor image is the contract boundary with the buf toolchain: the wire-stub templates consume it directly, and it is the only place the annotation options are legible (they are defined in the schema itself, so values decode dynamically against the image, and diagnostics carry its source positions). The emitters never see protos or descriptors; the normalized model is the only contract they know.

### Shared Semantics

All emitters use the same descriptor classification rules:

| Construct | SDK semantics |
| --- | --- |
| [Scalar fields](https://protobuf.dev/programming-guides/proto3/#scalar) | No presence; language-native defaults. Scalars marked [`optional`](https://protobuf.dev/programming-guides/proto3/#field-labels) have explicit presence. |
| [Message fields](https://protobuf.dev/programming-guides/proto3/#other) | Presence. |
| [`repeated` fields](https://protobuf.dev/programming-guides/proto3/#field-labels) | Default to empty collections. |
| [`map` fields](https://protobuf.dev/programming-guides/proto3/#maps) | Default to empty maps. |
| [`enum`](https://protobuf.dev/programming-guides/proto3/#enum) | Unknown numeric values are preserved. |
| [`oneof`](https://protobuf.dev/programming-guides/proto3/#oneof) | Explicit variants plus an unset state. |
| [`google.protobuf.Struct`](https://protobuf.dev/reference/protobuf/google.protobuf/#struct), [`Value`](https://protobuf.dev/reference/protobuf/google.protobuf/#value) | A JSON object and a JSON value respectively, with presence. |
| [`google.protobuf.Timestamp`](https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp) | A native timestamp, with presence. |
| [`google.protobuf.Duration`](https://protobuf.dev/reference/protobuf/google.protobuf/#duration) | A native duration, with presence. |
| [`google.protobuf.Empty`](https://protobuf.dev/reference/protobuf/google.protobuf/#empty) | As a request: a method with no public request argument. As a field: a unit variant. |
| [`google.rpc.Status`](https://github.com/googleapis/googleapis/blob/master/google/rpc/status.proto) | The canonical SDK error (see Errors), with presence. |

Language emitters own idiomatic syntax, naming, imports, and file layout. They must not reinterpret these semantic rules.

### SDK Annotations

The generator emits everything derivable from the descriptor, but the most ergonomic surfaces depend on facts that live in proto comments. A small Gestalt-owned annotation vocabulary moves them into the schema, where the generator acts on them once for all four languages.

This is the approach proven by [GAPIC](https://google.aip.dev/client-libraries/4210), whose generators derive flattened call signatures, pagination, and resource types from `google.api.*` options. GAPIC itself does not fit because it assumes APIs that follow Google's [AIP design standards](https://google.aip.dev/).

Annotations are a closed, validated vocabulary. Each is defined once as a language-neutral rule in the style of the Shared Semantics table, and emitters render it natively. The initial vocabulary:

| Annotation | Type | Meaning |
|---|---|---|
| `signature` | [`method`](https://protobuf.dev/reference/protobuf/google.protobuf/#method) | The listed request fields become parameters, in order ([AIP-4232](https://google.aip.dev/client-libraries/4232)). Presence-bearing fields come last and render as optional trailing parameters. |
| `optional_signature` | [`method`](https://protobuf.dev/reference/protobuf/google.protobuf/#method) | The listed request fields become a trailing bag of optional call options ([AIP-4232](https://google.aip.dev/client-libraries/4232)). |
| `initial` | [`method`](https://protobuf.dev/reference/protobuf/google.protobuf/#method) | The streamed message carries a oneof: the first message carries the named header variant, and every later message carries the named chunk variant. |
| `json_result` | [`method`](https://protobuf.dev/reference/protobuf/google.protobuf/#method) | The named bytes field of the result carries the standard JSON operation envelope; the method returns the decoded payload and raises the canonical invoke error on envelope or HTTP-error failures. |
| `optional_result` | `message` | The named value field becomes the optional return, present only when the named bool guard field is true. |
| `keyed` | `message` | The named repeated entries field becomes a native map from each entry's key field to its value field, omitting entries whose present field is false. |
| `unwrap` | `message` | The named field becomes the return value: a response whose only meaning is one field returns that field directly. |
| `host_binding` | `service` | The host-service binding name; generates a zero-configuration connect constructor that dials the binding advertised through the host environment. |

Annotation references are resolved against the descriptor, never trusted as strings: every named field, variant, and type is verified during model building, so renaming a schema element without moving its annotations fails generation, and CI's `sdkgen --check`, before any file is written. Diagnostics carry the source position of the declaration and the fully-qualified construct:

```
sdk/proto/v1/app.proto:307:3: method gestalt.provider.v1.AppProvider.Execute: signature names unknown request field "operaton"
```

Each annotation is rendered natively by every emitter; the examples below show TypeScript for brevity.

#### Examples

<details>
<summary><code>signature</code> + <code>optional_signature</code></summary>

```proto
rpc Invoke(AppInvokeRequest) returns (OperationResult) {
  option (signature) = "app";
  option (signature) = "operation";
  option (signature) = "params";
  option (optional_signature) = "connection";
  option (optional_signature) = "idempotency_key";
}
```

```ts
// without
const result = await app.invoke({ app: "github", operation: "get_issue", params, connection: "work" });

// with
const result = await app.invoke("github", "get_issue", params, { connection: "work" });
```

</details>

<details>
<summary><code>initial</code></summary>

```proto
rpc ReadObject(ReadObjectRequest) returns (stream ReadObjectChunk) {
  option (initial) = { header: "meta", chunk: "data" };
}
```

```ts
// without: a stream of frames, each carrying meta or data
for await (const frame of s3.readObjectRaw(request)) {
  if (frame.result.case === "meta") { /* first frame */ }
  if (frame.result.case === "data") { /* every later frame */ }
}

// with: the first-frame rule is generated
const { meta, data } = await s3.readObject(request);
const bytes = await data.transformToByteArray();
```

</details>

<details>
<summary><code>json_result</code></summary>

```proto
rpc Invoke(AppInvokeRequest) returns (OperationResult) {
  option (json_result) = { status: "status", body: "body" };
}
```

```ts
// without
const result = await app.invokeRaw({ app: "github", operation: "get_issue", params });
const issue = JSON.parse(new TextDecoder().decode(result.body)).data;

// with
const issue = await app.invoke("github", "get_issue", params);
```

</details>

<details>
<summary><code>optional_result</code></summary>

```proto
message CacheGetResponse {
  option (optional_result) = { guard: "found", value: "value" };
  bool found = 1;
  bytes value = 2;
}
```

```ts
// without
const response = await cache.get("session");
const value = response.found ? response.value : undefined;

// with
const value = await cache.get("session"); // Uint8Array | undefined
```

</details>

<details>
<summary><code>keyed</code></summary>

```proto
message CacheGetManyResponse {
  option (keyed) = { entries: "entries", key: "key", present: "found", value: "value" };
  repeated CacheEntry entries = 1;
}
```

```ts
// without
const response = await cache.getMany(keys);
const out: Record<string, Uint8Array> = {};
for (const entry of response.entries) {
  if (entry.found) out[entry.key] = entry.value;
}

// with
const out = await cache.getMany(keys); // { [key: string]: Uint8Array }
```

</details>

<details>
<summary><code>unwrap</code></summary>

```proto
message GetSecretResponse {
  option (unwrap) = "value";
  string value = 1;
}
```

```ts
// without
const secret = (await secrets.getSecret("name")).value;

// with
const secret = await secrets.getSecret("name"); // string
```

</details>

<details>
<summary><code>host_binding</code></summary>

```proto
service Cache {
  option (host_binding) = "cache";
}
```

```ts
// without
const transport = createHostServiceGrpcTransport(parseTarget(env.GESTALT_HOST_SERVICE_SOCKET), env.GESTALT_HOST_SERVICE_TOKEN, "cache");
const cache = new Cache(transport);

// with
const cache = Cache.connect();
```

</details>

### Generated and Handwritten Code

Generated code is the public SDK surface. Every generated file carries a standard generated-by header, is never hand-edited, and the generator deterministically owns its complete set of output files so stale outputs are removed on regeneration. `sdkgen` is the single entry point for generation; there is no wrapper script and there are no subcommands, because the tool does exactly one thing:

```bash
sdkgen [--target ts,python,go,rust] [--check]
```

The inputs are:

- The proto module at `sdk/proto`, compiled to a Buf descriptor image as the first step
- A pinned toolchain (buf and the per-language formatters)
- The generator-owned files already in the SDK trees

`--target` (`-t`) narrows emission to a subset of languages for local iteration and defaults to all four.

A plain run regenerates the wire stubs (still produced by the Buf templates, per the non-goals), emits and formats the SDK surfaces, and removes stale generator-owned files; it never touches handwritten files.

Handwritten files live alongside generated ones. They may wrap, extend, and re-export generated types to add convenience, and the generator never writes to or deletes them. Generated files carry doc comments derived from proto comments where available; narrative documentation stays in handwritten files. The handwritten layer is for one-off conventions such as the [IndexedDB transaction state machine](https://github.com/valon-technologies/gestalt/blob/ad4174c53f97c7a27e8f00f01cbd6a2dc2aa760a/sdk/typescript/src/indexeddb.ts#L981).

### CI Checks

With `--check`, the same rendering goes to a scratch location and is byte-compared against the tree without mutating anything, exiting nonzero with a per-file summary on drift; CI runs `sdkgen --check` on every change. Either way, an unsupported proto construct or a toolchain version mismatch fails with a clear diagnostic before any file is written.

hughhan1:
## Summary

Closes #2363.

Answers the deletion question the spike set out to settle: the handwritten cache client now delegates to the generated `Cache` client, dropping its hand-rolled transport construction, wire conversions, found/value flattening, and duration plumbing — 72 lines of plumbing become 14 lines of delegation, with `CacheRpc.connect(name)` replacing the manual host-service target parsing, transport assembly, and interceptor wiring. What survives is the provider half (server scaffolding is a later slice), the unchanged public interface, and a thin adapter; nothing about the module's public API moves.

The TypeScript suite passes unchanged, including the live cache-provider integration test now running end to end through the generated client. That test is what originally caught that keyed collapses must build null-prototype records so `__proto__` cache keys become data properties — a hardening detail the handwritten layer had, now encoded in the generator once for every language rather than once per hand-maintained client.

## Interfaces

```ts
class CacheImpl implements Cache {
  private readonly client: CacheRpc;

  constructor(name?: string) {
    this.client = CacheRpc.connect(name);
  }
  // public Cache interface unchanged
}
```

## Validation

- `tsc --noEmit` and eslint clean; the full TypeScript suite passes, including the live cache-provider integration test exercising get/set/getMany/touch through the generated client.
- The diff is confined to `sdk/typescript/src/providers/cache.ts`; no generated file changes, so `sdkgen --check` is unaffected.


hughhan1:
Superseded by #2383: the handwritten Cache facade this spike rewired has been deleted there, and the generated clients (with client-level ambient request context) are now the only client surface in all four SDKs.

PR #2395, opened 2026-06-10
Rename datastore proto surface to indexeddb

hughhan1:
## Summary

Renames the shared provider protocol file sdk/proto/v1/datastore.proto to indexeddb.proto and regenerates the Go, Python, TypeScript, and Rust bindings under the new name. The proto package already defined the IndexedDB service, and every user-facing surface (manifest kind, config keys, the PROVIDER_KIND_INDEXEDDB wire enum) already said indexeddb; this aligns the file name, generated modules, and remaining in-repo naming with that convention.

Beyond the generated code, the rename covers identifiers and user-facing strings that still said datastore: gestaltd config validation and readiness naming, bootstrap and provider-driver error messages, and the routinggen go:generate directive. The OpenTelemetry GenAI tool-type constants keep the "datastore" value because the gen_ai.tool.type semantic convention defines it, and generic docs prose is unchanged.

## Interfaces

```go
// sdk/go: the ProviderKind value now matches the manifest kind.
ProviderKindIndexedDB ProviderKind = "indexeddb" // was "datastore"

// OpenDB reads the renamed env var.
// GESTALT_INDEXEDDB_<ALIAS> (was GESTALT_DATASTORE_<ALIAS>)
```

```python
from gestalt._gen.v1 import indexeddb_pb2, indexeddb_pb2_grpc  # was datastore_pb2*
```

```ts
import { IndexedDB } from "./internal/gen/v1/indexeddb_pb.ts"; // was datastore_pb.ts
```

## Runtime

The provider runtime handshake is unchanged on the wire: ProviderKindIndexedDB is converted to the existing PROVIDER_KIND_INDEXEDDB enum before transport. The GESTALT_INDEXEDDB_<ALIAS> env var rename is a breaking change for deployments that configure DSNs through gestalt.OpenDB; nothing in this repository or its sibling repositories sets the old variable. Readiness probe failure text now reports "indexeddb unavailable" instead of "datastore unavailable".


hughhan1 (commented):
Automated code review (line-by-line diff scan, removed-behavior audit, cross-file caller trace, plus reuse/simplification/efficiency/altitude passes): no actionable findings.

Verified: all callers of renamed identifiers (`validateIndexedDBConfig`, `indexedDBPinger`, `readinessIndexedDBPingTimeout`, test helpers) are updated; the regenerated bindings are referenced consistently across Go/Python/TypeScript imports and the `routinggen` go:generate directive; `ProviderKindIndexedDB` is converted to the unchanged `PROVIDER_KIND_INDEXEDDB` wire enum, and no code matches on the old "datastore" string; no in-repo reader of `GESTALT_DATASTORE_*` remains; the Rust generated output is byte-identical after regeneration.

Two deliberate exclusions, called out in the PR description: the `GENAI_TOOL_TYPE_DATASTORE` constants keep the "datastore" value because that is the OpenTelemetry `gen_ai.tool.type` semantic-convention value, and the `GESTALT_INDEXEDDB_<ALIAS>` env var rename is an accepted break for external deployments (nothing in this repo or sibling repos sets the old name). One considered-and-skipped cleanup: `runtime.go`'s local `indexedDBPinger` interface duplicates `core/indexeddb.Pinger`, but the consumer-side minimal interface is idiomatic Go and predates this PR.

PR #2400, opened 2026-06-11
Mark the registry app with a stable data attribute

gvocale:
## Summary

The routing check scripts decide "this response is the registry app" by searching served HTML for `registry_shell__` — the hashed CSS-module class prefix the build happens to emit. That's an accident of the build, not a contract: moving or renaming `registry.module.css` (which the registry rehome later in this stack does), or any change to Next's module hashing, silently breaks every assertion.

This replaces the marker with an explicit `data-registry-shell="true"` attribute on the registry's root `<main>`, and points all 14 assertion sites across `check-nginx-routing.mjs`, `check-registry-static.mjs`, and `check-registry-docs-render.mjs` at it. Landing the swap first keeps the later PRs' diffs free of this noise.

### DIFF
--- a/docs/app/registry/registry-app.tsx
+++ b/docs/app/registry/registry-app.tsx
@@ -251,7 +251,7 @@ export default function RegistryApp({ catalogUrl }: { catalogUrl: string }) {
       : `${styles.shell} ${styles.lightShell}`;
 
   return (
-    <main className={shellClassName}>
+    <main className={shellClassName} data-registry-shell="true">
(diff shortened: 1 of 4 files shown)


hughhan1 (approved):

gvocale:
Reopening was blocked (head was force-pushed while closed — GitHub closed this PR when its base branch was deleted during the stack merge). Continued as a fresh PR with identical content; needs re-approval there.

PR #2404, opened 2026-06-11
Serve the provider index at registry.gestaltd.ai

gvocale:
## Summary

`https://registry.gestaltd.ai/provider-index.yaml` — the URL our docs tell people to `gestaltd provider repo add` — now actually serves the provider index. That path has always fallen through to the registry SPA's HTML; anyone running the documented command got a parse failure. Gestalt is open source and consumers will run their own provider repositories, so the canonical example stays on the project-neutral domain — and now it works.

The index is a published artifact of the registry deployment: baked into the image at build (`ADD` from the gestalt-providers repo), served by an exact-match nginx location as `application/yaml`, asserted in the docker-based routing check (`schema: gestaltd-provider-index`). A runtime proxy to GitHub was rejected — it would let the browsable registry and the machine index drift apart. The two fetches (site build vs image build) are minutes apart on the same pipeline; pinning both to one ref is a follow-up if it ever bites.

The example alias also changes from `valon` to `gestalt`.

## Demo

provider index

[Full-quality video](https://github.com/valon-technologies/gestalt/blob/pr-media/registry-provider-index-2026-06-11.mov)


hughhan1 (approved):

gvocale:
bugbot run

PR #2405, opened 2026-06-11
Give gestalt its own branding with free fonts

gvocale:
## Summary

Gestalt's docs get their own identity, independent of Valon: **silver neutrals with the gold accent** (matching the favicon mark), **Instrument Sans** for UI/body, **Newsreader at optical size 72** for display headings, Geist Mono for code — all OFL, vendored with license texts. The Valon-era palette values are swapped in place inside `shell.css`'s token block, and the licensed KMR Melange Grotesk / Season Serif binaries **leave the repo**.

## ⚠️ Licensing

The committed brand woffs are commercial retail fonts; Season Serif's embedded license forbids public-server storage outright. They've been world-downloadable since April — merging ends the ongoing distribution at HEAD (tags keep them; a history purge would break versioned builds, separate decision).

Font picks were measured against the real woffs: Instrument Sans matches KMR's x-height exactly (body tracks the variable axis at 435 to match its darker regular); Newsreader-72 restores Season Serif's line length where the prior pick ran −22% condensed. Old tags' baked font rules are beaten by working-tree re-declarations; lazy `@font-face` means removed binaries are never requested.

Also: the image build no longer exports layers to the public Actions cache (fork-restorable; never hit anyway).

**Merging changes the deployed look — that's the point.**

## Demo

gestalt brand

[Full-quality video](https://github.com/valon-technologies/gestalt/blob/pr-media/docs-gestalt-brand-2026-06-11.mov)


hughhan1 (approved):

gvocale:
bugbot run

PR #2429, opened 2026-06-12
[provider-gateway] split transport dependency

mzywang:
## Description

Splits the provider gateway transport interface from the full gateway so authorization providers can depend on direct transport while the full gateway can be wired with the selected authorization provider after bootstrap.


mzywang:

- #2429 ◀
    - #2461

PR #2432, opened 2026-06-12
sdkgen: generate provider-side surfaces for all unary services (Go)

hughhan1:
## Summary

Generates the provider-implementation (server-side) SDK surface from the protos for **every unary provider service**, and routes the root package's provider serving through it. A new `provider` service annotation marks "providers implement this service"; for each marked service the Go emitter generates the native handler interface with Unimplemented defaults and a wire dispatch adapter built on the existing bidirectional codecs. Annotated: `Workflow`, `Agent`, `AppProvider`, `Authentication`, `Authorization`, `Cache`, `ExternalCredentials`, `Runtime`, `Secrets`. S3 and IndexedDB stay handwritten until the emitters learn streaming dispatch (validation rejects `provider` on streaming services).

Two emitter changes unblock the convergence: generated native fields use Go initialism spellings (`RunID`, `MetadataJSON` — matching the handwritten surface; wire-stub names keep protoc style), and native message types gain nil-safe getters in the handwritten surface's exact convention.

The root package now serves **all eight unary provider kinds through the generated adapters**: facade kinds (secrets, cache) via thin native-only bridges, twin kinds (workflow, agent, authentication, authorization, runtime, external credentials) via bridges converting root natives ↔ generated natives — deleting every handwritten proto-conversion path on the serve side (~2,300 lines of wire plumbing removed; authorization's and agent's proto converter files swept entirely). A shared unary interceptor scopes the inbound `RequestContext` for every kind. The root provider interfaces, types, and `Serve*` signatures are unchanged — all eight downstream gestalt-providers Go modules compile against this SDK without edits (verified via workspace replace).

## Interfaces

```proto
extend google.protobuf.ServiceOptions {
  string host_binding = 50001;  // existing: zero-config client constructor
  bool provider = 50002;        // NEW: generate the provider handler surface and dispatch adapter
}
```

```go
// Generated per provider service into sdk/go/client/<base>_server.go:
type WorkflowProvider interface {
    ApplyDefinition(ctx context.Context, request *ApplyWorkflowProviderDefinitionRequest) (*WorkflowDefinition, error)
    // ... one method per RPC; Empty responses collapse to error-only
}
type UnimplementedWorkflowProvider struct{}
func NewWorkflowProviderServer(provider WorkflowProvider) proto.WorkflowServer

// Generated natives now carry initialism field names and nil-safe getters:
run.GetRunID(); definition.GetCreatedBySubjectID()
```

Generated client field renames are monorepo-internal (`sdk/go/client` consumers and runtime-compiled gestaltd fixtures updated here, including the fixture program embedded as a string in testutil). The handler interface for a service already named `*Provider` takes a `Handler` suffix (`AppProviderHandler`) since its generated client owns the bare name. No exported root-package names changed.

## Runtime

Wire behavior is preserved end to end: bridges call `providerRPCError` before the generated status mapping, so sentinel errors (`ErrSecretNotFound`, `ErrExternalCredentialNotFound`, …) keep their codes and operation strings; the workflow bridge keeps `DeliverEvent`'s nil-result echo fallback and the `valueFromAny`/`structFromAny` validation errors; agent keeps `TrimSpace` normalization and the turn-output oneof mapping. Every provider transport test passes unchanged — they gate the swap. A new transport test additionally round-trips the fully generated loop (generated client ↔ generated adapter ↔ native handler). All gates green: sdkgen tests + `--check` across four targets, full sdk/go suite, golangci-lint 0 issues, gestaltd `go test ./... -run '^$'` (compiles the runtime-built fixtures), GOWORK=off fixture builds, and all eight downstream provider modules.

The remaining step beyond this PR is deleting the root twin types in favor of the generated natives (`gestalt.WorkflowRun` *becoming* `client.WorkflowRun`) — a breaking change for provider authors (pointer slices, `time.Time` handling, `Subject` vs `SubjectContext`) that needs a coordinated downstream migration; the bridges localize exactly the code that disappears when it lands.


hughhan1:
Superseding this in favor of a horizontal PR stack per the re-slicing decision: the work is split by layer (annotation → per-language emitter → per-language emit+delete → per-language downstream migration) and converges fully (deleting the twin types) rather than bridging. Foundation PR: #2433. The Go emitter + getter-lean rendering proven here are reused; the bridge-based Go convergence is replaced by full twin deletion in its own PR.

Closing to keep the stack clean.

PR #2474, opened 2026-06-16
Fix local gestaltd bring-up

hughhan1:


hughhan1:
Superseded by aligning provider snapshots to the from-source gestaltd binary and merging #2464 into PR #2472. No gestaltd bring-up workarounds needed.

PR #2492, opened 2026-06-20
[http] expose binding raw body to operations

mzywang:
## Description

Expose hosted HTTP binding request bodies as base64 workflow context so webhook handlers can verify signatures against the exact raw bytes received.


mzywang:

- #2492 ◀

PR #2493, opened 2026-06-22
[providergateway] tag operation.count with invocation surface

mzywang:
## Description

Adds a `gd.invocation_surface` tag to `gestaltd.provider_gateway.operation.count` so operations can be attributed to their originating surface (`http`, `http_binding`, `mcp`, app-access). The existing `gd.source` tag only distinguishes `sdk_grpc`/`internal` and cannot tell these apart.

The surface is already present in `ctx` at record time (every surface calls `invocation.WithInvocationSurface` before invoking), so the recorder just reads `invocation.InvocationSurfaceFromContext(ctx)`. Paths that never set a surface fall back to `unknown` via `metricutil.AttrValue`. Verified no import cycle (`invocation` does not depend on `providergateway`).

### DIFF
--- a/gestaltd/services/providergateway/metrics.go
+++ b/gestaltd/services/providergateway/metrics.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	proto "github.com/valon-technologies/gestalt/server/rpc/protov1/v1"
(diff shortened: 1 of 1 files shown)


mzywang:

- #2493 ◀

PR #2506, opened 2026-06-23
[packageio] buffer archive reads to cut syscall overhead

mzywang:
walkPackageArchive fed the raw *os.File straight to gzip.NewReader, so the
flate decompressor pulled its input a few bytes at a time, issuing one read()
syscall per few bytes. Wrap the file in a 64KB bufio.Reader so decompression
reads in large blocks.

Profiling the operator race tests (which exercise this path heavily) showed
~79% of wall time in syscall.read before this change; it drops to ~0 after.
Aggregate user CPU for the operator package under -race fell ~21% (615s->488s).
Pure I/O optimization, no behavior change.

### DIFF
--- a/gestaltd/services/apps/packageio/archive.go
+++ b/gestaltd/services/apps/packageio/archive.go
@@ -2,6 +2,7 @@ package packageio
 
 import (
 	"archive/tar"
+	"bufio"
 	"compress/gzip"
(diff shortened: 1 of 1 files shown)


mzywang:

- #2506 ◀

PR #2507, opened 2026-06-23
[operator] build shared Go secrets provider binary once

mzywang:
buildGoSourceSecretsBinary compiled a byte-identical Go provider binary on
every call (5 call sites across the source-secrets tests), each a full `go build`
of an SDK-importing main. Under -race these builds were a meaningful chunk of the
operator race job's wall-clock.

Compile the binary once per test-binary run via sync.Once into a shared temp
directory and reuse it; the directory is removed in TestMain. The binary is
identical for all callers and is consumed read-only (callers either read its
bytes or pass its path), so sharing is safe. Verified under -race -parallel.


mzywang:

- #2507 ◀

PR #2508, opened 2026-06-23
[packageio] stop decompressing provider archives redundantly

mzywang:
Installing a provider package decompressed the same gzip/tar archive three
times: once to read the manifest, once to read the artifact for its digest
check, and once to extract. Manifest resolution also walked the whole archive
once per candidate name, so packages using manifest.yaml/.yml paid an extra
full decompression for the manifest.json miss.

- ReadPackageManifestIn now collects whichever candidate manifest is present in
  a single walk instead of one full walk per name.
- installPackage extracts first, then verifies the artifact digest against the
  extracted file (via packageio.FileSHA256) instead of decompressing the archive
  a second time to read it. The package is never executed before the check.

Full-archive duplicate-entry detection (walkPackageArchive) and single-root
manifest validation (ExtractPackage) are unchanged. Measured on the operator
race suite, which exercises this path heavily: package user CPU under -race fell
~54% (615s -> 285s).

### DIFF
--- a/gestaltd/internal/operator/package_install.go
+++ b/gestaltd/internal/operator/package_install.go
@@ -1,8 +1,6 @@
 package operator
 
 import (
-	"crypto/sha256"
-	"encoding/hex"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


mzywang:

- #2508 ◀

PR #2511, opened 2026-06-23
Forward provider process logs to Cloud Logging

hughhan1:
## Summary

- Tee provider stdout/stderr to the container stderr alongside the existing in-memory runtime session log store, so provider logs are visible in Cloud Logging.
- Add a `providerLogWriter` that prefixes each line with `provider=<name>` so provider output is distinguishable from host logs in Cloud Logging queries.
- Buffer incomplete lines and flush them on newline or `Close`; wire `Close` into `ProcessConfig.Cleanup` so the final partial line is emitted when the process shuts down.
- Keep the session log store unchanged; the tee is additive and the non-session-log path is unaffected.
- Verify `runtimeprovider/executable.go` does not override stdout/stderr, so it continues to default to `os.Stderr` via `runtimehost.StartAppProcess`.

## Test plan

- [x] Unit test: `TestLocalProviderTeesSessionLogsToStderr` asserts that fake provider output appears in both the session log store and the captured stderr with the `provider=` prefix.
- [x] Unit test: `TestLocalProviderNoSessionLogsDefaultsToStderr` asserts that without session logs, stdout/stderr still reach `os.Stderr` without any prefix and without session log capture.
- [x] Unit tests: `TestProviderLogWriter*` cover complete lines, buffered partial lines, flush on close, and empty-name fallback.
- [x] Package tests: `go test ./gestaltd/services/runtimehost/...` passes.
- [ ] Manual verification after deploy: open a PR against `valon-technologies/toolshed` and confirm `provider=codeReview` lines appear in `run.googleapis.com/stderr`.

### DIFF
--- a/gestaltd/services/runtimehost/runtimeprovider/local.go
+++ b/gestaltd/services/runtimehost/runtimeprovider/local.go
@@ -315,9 +315,23 @@ func (p *LocalProvider) StartApp(ctx context.Context, req *proto.StartHostedAppR
 
 	stdout := io.Writer(nil)
 	stderr := io.Writer(nil)
+	var processCleanup func()
 	if p.sessionLogs != nil {
-		stdout = newSessionLogWriter(p.sessionLogs, p.runtimeProviderName, req.GetSessionId(), runtimelogs.StreamStdout, &session.logSeq)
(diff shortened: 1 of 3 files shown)


hughhan1:
Closing per request.

PR #2517, opened 2026-06-23
Skip gestalt-managed dirs in source fingerprints

gvocale:
## What

Source-input fingerprints now skip gestalt's own managed directories — `.gestaltd` (artifacts/state) and `.gestalt` (build scratch) — the same way they already skip `.git`.

## Why (in simple terms)

When you run an isolated local-dev backend with `--artifacts-dir` placed **inside** a provider's source tree, `gestaltd sync` could never finish: it kept reporting the artifact it had **just built** as "missing or stale," looping forever.

The cause: to decide whether a prepared artifact is still fresh, gestaltd fingerprints the provider's **source** files. But that walk descended into the artifacts directory when it was nested under the source. Materializing an artifact renames a staging dir (`X.tmp-123` → `X`) and writes a lock-metadata file — so the "source" fingerprint changed **between** the moment gestaltd recorded it and the moment it re-checked it. The re-check saw a different digest and declared the just-built artifact stale.

`sync` (which validates source inputs) hit this on the first provider and never converged. `serve --locked` (which skips source-input validation) got one provider further before failing the same way — so the two commands **disagreed about the same artifact**.

The fix treats gestalt's artifact/build directories as non-source, exactly like `.git`. They are **outputs**, never source inputs, so excluding them keeps the source fingerprint stable no matter where `--artifacts-dir` lives.

## Test

`TestFingerprintExcludesNestedGestaltManagedDirs` reproduces the exact churn (a nested `.gestaltd` artifacts dir + staging→final rename + lock-metadata write) and asserts the source fingerprint stays stable. It fails without the fix and passes with it.

### DIFF
--- a/gestaltd/internal/operator/fingerprint.go
+++ b/gestaltd/internal/operator/fingerprint.go
@@ -15,16 +15,40 @@ import (
 )
 
 // fingerprintSkipper is the fingerprint exclusion policy: which paths a source
-// fingerprint walk excludes. Two structural invariants apply regardless of
+// fingerprint walk excludes. Structural invariants apply regardless of
 // .gitignore: the declared build output (never hash the artifact about to be
-// produced) and the .git directory itself (git metadata, never a build input).
-// Everything else is decided by the enclosing repo's .gitignore rules. Explicit
-// per-input file listings are honored verbatim by the callers; shouldSkip is
-// only consulted on directory walks and on the walk root.
+// produced), the .git directory (git metadata, never a build input), and
+// gestalt's own managed directories (see reservedFingerprintDir). Everything
+// else is decided by the enclosing repo's .gitignore rules. Explicit per-input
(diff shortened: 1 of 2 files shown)


gvocale:
Closing: not needed for the local-dev use case it was written for.

origin/main #2505 made source fingerprinting `.gitignore`-aware, and the `/local-dev` skill's `allocate` writes an app `.gitignore` (`.local/ .gestaltd/ .gestalt/`) and places `--artifacts-dir` under gitignored `.local/`. So the nested artifacts dir is already excluded from the source fingerprint on a plain origin/main binary — `sync` converges without this structural skip.

The structural-skip idea still has standalone merit (a guarantee independent of each consumer's `.gitignore`, and it covers `.gestalt` scratch which isn't always gitignored). Reopen if we want that defense-in-depth upstream.

PR #2525, opened 2026-06-23
Classify app releases with platform artifacts as executable

hughhan1:
## Summary
- App providers with both a spec surface (OpenAPI/rest) and a build that produces a hosted binary were misclassified as `runtime: declarative` in `provider-release.yaml`.
- The packaging stage correctly sets `Entrypoint` and `Artifacts` on the prepared manifest, but `staticvalidation.ProjectManifest` strips both for the platform-neutral validation manifest. `RuntimeForManifest` then saw `Entrypoint == nil` with `IsManifestBacked() == true` and returned `RuntimeDeclarative`, even though the release contained platform-specific binaries.
- This caused gmail, bigquery, hex, and vercel to publish with `runtime: declarative` and an empty catalog fingerprint, so their custom Python operations were never served at runtime.

## Key changes
- Add `RuntimeForRelease(kind, manifest, artifacts)` in `providerrelease/release.go` — classifies based on both the validation manifest and published artifact targets. A release with platform-specific artifacts is `executable` regardless of the stripped validation manifest.
- Update `buildProviderReleaseMetadata` to set `Runtime` via `RuntimeForRelease` after populating the artifacts map.
- Update `ValidateMetadata` to validate runtime via `RuntimeForRelease` instead of `RuntimeForManifest`.

## Interfaces
- New exported function `providerrelease.RuntimeForRelease`. `RuntimeForManifest` is unchanged.

## Test plan
- [x] `TestProviderReleaseMetadataClassifiesHybridAppAsExecutable` — app with OpenAPI surface + entrypoint + platform artifacts → `executable`
- [x] `TestProviderReleaseMetadataClassifiesSpecOnlyAppAsDeclarative` — app with OpenAPI surface, no entrypoint, generic artifact → `declarative`
- [x] Full `gestaltd/internal/daemon` and `gestaltd/internal/providerrelease` test suites pass
- [ ] Re-publish gmail/bigquery/hex/vercel snapshots and verify `runtime: executable` in `provider-release.yaml`
- [ ] Repin deploy lock file and verify non-empty catalog fingerprint

### DIFF
--- a/gestaltd/internal/daemon/provider_release_finalize.go
+++ b/gestaltd/internal/daemon/provider_release_finalize.go
@@ -252,7 +252,6 @@ func buildProviderReleaseMetadata(manifest *providermanifestv1.Manifest, version
 		Package:       manifest.Source,
 		Kind:          manifest.Kind,
 		Version:       version,
-		Runtime:       providerrelease.RuntimeForManifest(manifest.Kind, staticManifest),
 		Artifacts:     providerrelease.Artifacts{},
 		StaticValidation: &providerrelease.StaticValidation{
 			Manifest: staticManifest,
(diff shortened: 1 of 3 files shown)


hughhan1:
Closing as duplicate of #2524, which already merged as d30206c8a with the same fix.

PR #2535, opened 2026-06-25
[gestalt] remove default access policy

mzywang:
## Description
Remove `defaultAccessPolicy` from authorization config, proto, and SDK surfaces. Mounted UI fallback now uses only explicit `defaultRole`, and the config API version is bumped to `gestaltd.config/v8` because this removes an accepted v7 field.

## Testing
- `go run ./sdk/tools/sdkgen --check` with buf v1.66.1
- `go test ./...` from `gestaltd`
- `go test ./...` from `sdk/go`
- `bun run typecheck` from `sdk/typescript`
- `uv run pytest` from `sdk/python`
- `cargo test` from `sdk/rust`


mzywang:

- #2535 ◀

PR #2545, opened 2026-06-26
[gestaltd] build the CI image once instead of twice

mzywang:
## Description

`build-gestaltd.yml` compiled gestaltd **twice** per main/dispatch run: a `Docker Image Build` job built the `static` + `alpine` targets and discarded them, and a separate `Publish CI Image` job rebuilt `static` from scratch (different runner, no shared output) and pushed it.

This collapses both into a single **`Build CI Image`** job that builds once and pushes conditionally:

- Builds the `static` image on every validating run (incl. PRs); pushes to GCP Artifact Registry **only** when the run may publish (`publish == 'true' && GESTALTD_CI_IMAGE_PUBLISH_ENABLED == 'true'`). On PRs it builds without pushing to validate the Dockerfile.
- Keeps `alpine`-target validation as a build-only step that reuses the compiled `build-binary` layer from the `static` build (a few seconds).
- Preserves existing publish behavior: immutable per-commit tag, skip-if-already-present, SBOM + provenance attestations, version smoke test, step summary. Registry auth + attestations are gated to publishing runs.
- Switches the BuildKit layer cache to `mode=max` so the compile layer is reused across runs.

### Behavior notes

- **Push safety is unchanged:** the job still `needs` the validation jobs (`lint`, `test`, `coverage`, `race`, `helm-validate`, `docs-build`), so an unvalidated commit is never published.
- **Trade-off:** Docker build feedback on PRs now runs after the other validation jobs instead of in parallel. Neither removed job was a required status check (required checks: Go Lint & Vet, Go Tests, Docs Build, Helm Chart Validation, Cursor Bugbot), so branch protection is unaffected.

### DIFF
--- a/.github/workflows/build-gestaltd.yml
+++ b/.github/workflows/build-gestaltd.yml
@@ -467,39 +467,6 @@ jobs:
         if: ${{ needs.resolve-ref.outputs.should_validate == 'true' }}
         run: kubeconform -strict -summary /tmp/rendered-preparation.yaml
 
-  docker-build:
-    name: Docker Image Build
-    needs: resolve-ref
-    if: ${{ needs.resolve-ref.outputs.should_validate == 'true' }}
-    runs-on: ubuntu-latest
-    timeout-minutes: 20
-    steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
-        with:
-          ref: ${{ needs.resolve-ref.outputs.sha }}
-
-      - name: Build static image
-        env:
-          DOCKER_BUILDKIT: "1"
-        run: |
-          docker build \
-            --file gestaltd/Dockerfile \
-            --target static \
-            --build-arg GESTALT_VERSION=ci \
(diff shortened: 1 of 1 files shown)


mzywang:

- #2545 ◀

mzywang:
Good catch — fixed. You're right that `build-image` needed `docs-build`, which is skipped on non-publishing runs, and GitHub propagates that skip to dependents. The same applies to `race` (skipped on pull requests), so PR runs would have skipped the Docker build entirely.

The original `publish-ci-image` avoided this only because its `if` was identical to `docs-build`'s, so they were always skipped/run together; merging into a job that also runs on PRs broke that coupling.

Reworked the gating to use `!cancelled()` with explicit `needs.*.result` checks: the always-run validations (`resolve-ref`, `lint`, `test`, `helm-validate`) must succeed, and the conditionally-skipped jobs (`coverage`, `race`, `docs-build`) are accepted when skipped but still required to succeed when they actually run. So the Docker build now runs on PRs, and publish-safety on main is preserved (race/docs-build must pass when present).

PR #2550, opened 2026-06-26
[gestaltd] build docs after publishing the CI image in CD

mzywang:
## Description

In `cd.yml`, `docs-build` previously ran before (and gated) the image push — `build-and-push` had `needs: [validate, docs-build]`, so a docs failure would block publishing the CI image. This reorders the pipeline so the image is built and pushed first, then docs build:

- `build-and-push`: `needs: validate` only; the gate simplifies to `should-validate == 'true'` (the `!cancelled()` + docs-build skip-tolerance is no longer needed).
- `docs-build`: `needs: [validate, build-and-push]`, runs after the image is published, and is skipped if the image build didn't succeed (default needs gating).

Net effect: `validate → build-and-push → docs-build`, and the image publish is no longer blocked by docs failures.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,10 +1,10 @@
 name: CD (gestaltd)
 
 # Continuous delivery for gestaltd. Re-runs the validation suite (including the
-# race matrix), builds the docs, and builds + pushes the per-commit CI image to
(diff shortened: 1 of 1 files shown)


mzywang:

- #2550 ◀

PR #2551, opened 2026-06-26
[gestaltd] don't run the full Go suite twice on main

mzywang:
The CD/scheduled validation runs both the Go Tests job (go test ./...) and the
Go Coverage job (go test -coverprofile ./...), which is the same full suite
executed twice. Coverage is a strict superset and already fails on any test
failure, so the plain run adds no signal.

Skip the Go Tests job whenever coverage is enabled (main push, workflow
dispatch, hourly schedule) and let coverage be the authoritative full-suite
gate. Pull requests keep the fast, package-scoped Go Tests job (coverage stays
off there). Add the Helm setup to the coverage job so it is a complete
replacement for tests that shell out to the helm binary.

### DIFF
--- a/.github/workflows/validate-gestaltd.yml
+++ b/.github/workflows/validate-gestaltd.yml
@@ -259,6 +259,7 @@ jobs:
   test:
     name: Go Tests
     needs: resolve-ref
+    if: ${{ !inputs.run-coverage }}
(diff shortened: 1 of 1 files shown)


mzywang:

- #2551 ◀
    - #2552
        - #2553

PR #2552, opened 2026-06-26
[gestaltd] compile-check PRs with go vet instead of go test -run '^$'

mzywang:
## Description

The PR compile-check ran `go test -run '^$' ./...` to confirm every package still builds even when only the affected packages' tests run. That executes each package's `TestMain` (`m.Run` with no tests), so the daemon/bootstrap/operator integration packages spent ~25s each building gestaltd + provider fixtures during a pass that runs nothing.

Switch the compile-check to `go vet ./...`: it type-checks and compiles every package including test files, but never executes `TestMain`, so those fixture builds no longer happen during the compile-check. The affected packages' tests still run afterward, and the full `go test ./...` on main/coverage continues to build and link every test binary.


mzywang:

- #2552 ◀
    - #2553

PR #2560, opened 2026-06-26
[gestaltd] build CI image from natively-compiled binary

mzywang:
## Description
Compile gestaltd on the runner with `actions/setup-go` (whose module + build caches persist across runs) and bake the binary into the image via the existing `static-prebuilt` / `alpine-prebuilt` Dockerfile targets, instead of running `COPY . . && go build` inside the Docker build.

The per-commit CI image changes its source every push, so the in-Docker build layer cache-missed on essentially every run and recompiled from scratch. Moving the compile out lets setup-go cache incrementally (only changed packages rebuild) and reduces the image build to a `COPY`. Reuses the same packaging/targets as release-gestaltd.yml. Verified locally end-to-end: native build → tarball → `static-prebuilt` and `alpine-prebuilt` images both build and `gestaltd version` reports the stamped version.

### DIFF
--- a/.github/actions/build-gestaltd-image/action.yml
+++ b/.github/actions/build-gestaltd-image/action.yml
@@ -1,11 +1,12 @@
 name: Build gestaltd CI image
 description: >-
-  Build the static gestaltd image and optionally push it to Artifact Registry.
-  When push is false the image is built (and discarded) to validate the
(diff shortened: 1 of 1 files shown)


mzywang:

- #2560 ◀

PR #2562, opened 2026-06-26
[ci] run docs build only on the hourly schedule

mzywang:
## Description
Move the docs build smoke check out of the CD path and onto ci.yml's hourly schedule. Adds a schedule-gated `docs-build` job to ci.yml (same pattern as `run-race`/`run-e2e`), removes the `docs-build` job from cd.yml, and strips build-docs.yml's own `push`/`pull_request` triggers so it runs only on the hourly schedule (or manual `workflow_dispatch`).

The docs build runs a full Next.js production build per version (~3-4 min) and its output is discarded — actual docs publishing is handled separately by deploy-gestalt-docs.yml. So it was pure validation that was slowing every gestaltd main push for no publish benefit.

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,17 +1,10 @@
 name: Build Docs
 
+# The docs build is an expensive, multi-build smoke check (it runs a full
+# Next.js production build per docs version). It is not run on pushes or pull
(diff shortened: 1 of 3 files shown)


mzywang:

- #2562 ◀

PR #2566, opened 2026-06-27
[gestaltd] grant the valon-tools GKE node SA read on the chart repo

mzywang:
gestaltd pods now pull the runtime image from gestaltd-charts (the
gestaltd-image sub-path) in gitlab-peach-street, a different project than
the cluster. The node SA's default role only grants Artifact Registry read
within its own project, so add the dev cluster node SA to the chart-reader
set; otherwise pods ImagePullBackOff at startup.

Stage: its cluster is not in valon-tools-infra, so its node SA
(tools-stage-nodes@valon-tools-stage by convention) should be added once
confirmed.


mzywang:

- #2566 ◀

PR #2570, opened 2026-06-27
[gestaltd] run publish jobs in parallel in CD

mzywang:
## Description

Fan out `build-and-push` (Publish Docker Image), `publish-alpine-image`, and `publish-chart` from `validate` instead of chaining them. The alpine image and Helm chart consume no output from the docker image publish, so gating them on it only added wall-clock latency. `publish-chart` no longer waits on the image jobs — the trade-off is a chart can publish without its companion images on a failed build.


mzywang:

- #2570 ◀

PR #2571, opened 2026-06-27
[gestaltd] build the alpine release image early in CD

mzywang:
## Description

Extract the multi-arch alpine release build into a composite action and add a `build-alpine-image` job that pushes the immutable `sha-<commit>-alpine` tag in parallel with `build-image` and the validation suite. `publish-alpine-image` now calls the same action and short-circuits on the existing tag, mirroring the `build-image` / `build-and-push` pattern — an early warm-build plus a post-validation fallback that retries the push if the early one failed. This overlaps the slowest publish step (arm64 via QEMU) with the test suite.


mzywang:

- #2571 ◀

PR #2576, opened 2026-06-27
[gestaltd] fail on orphaned checksum in binary publish

mzywang:
## Description

Follow-up to #2575. The create-only binary upload logic mishandled one residual inconsistent state: if a `.tar.gz.sha256` exists in GCS but its tarball is missing, the job rebuilt and uploaded a fresh archive (different gzip mtime → different bytes) yet skipped the checksum upload because `has_sha` was already true — leaving the published checksum disagreeing with the new tarball, and `sha256sum -c` would fail for consumers.

Since `objectCreator` cannot overwrite the stale checksum to repair it, this orphaned state is now a hard error. Past that guard `has_sha` is always false, so the checksum is always derived from the authoritative tarball bytes (downloaded if already published, freshly built otherwise) and uploaded unconditionally.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -223,6 +223,18 @@ jobs:
               continue
             fi
 
+            # A published checksum with no tarball is an orphaned, inconsistent
+            # state: we cannot republish a matching tarball without overwriting the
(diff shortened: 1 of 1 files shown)


mzywang:

- #2576 ◀

PR #2577, opened 2026-06-27
[gestaltd] build catalog binary for host libc (fix bun app publish on glibc)

mzywang:
## Description

`gestaltd provider package` generates the static catalog by exec'ing the compiled provider binary. For TypeScript providers the SDK always compiled `bun-linux-x64-musl`, so on a **glibc** runner (GitHub `ubuntu-latest`, where Valon Tools app snapshots are published) the catalog exec failed with `no such file or directory` — the musl ELF loader `/lib/ld-musl-x86_64.so.1` is absent. This broke publishing for **every bun app** (g-issues, etc.). The deploy path was unaffected because it builds on alpine/musl.

Completes the existing-but-unwired `GESTALT_TARGET_LIBC` / `SourceBuildOptions.LibC` plumbing:
- **SDK** (`sdk/typescript/src/build.ts`): `bunTarget` honors libc — `glibc`/`gnu` → `bun-linux-x64`; default stays musl (no change for existing callers).
- **gestaltd** (`providerpkg`): the host-side catalog build targets the **build host's libc** (`HostLibC`) so it execs locally; the shipped artifact stays **musl**. The catalog build is reused as the artifact only when platform **and** libc match, so a glibc host catalog build is followed by a musl target build for the deploy runtime.

## Verification
- SDK target mapping: `linux/amd64` default → `bun-linux-x64-musl` (unchanged), `glibc` → `bun-linux-x64`, non-linux ignores libc.
- `go test ./services/apps/providerpkg/...` passes.
- Darwin packaging still emits a musl `linux/amd64` artifact (deploy unaffected).
- In a glibc container the catalog `ENOENT` is gone (reproduced before; resolved after).

## Rollout (both halves required to take effect)
1. Publish a new `@valon-technologies/gestalt` SDK including the `build.ts` change and bump g-issues' (and other bun apps') dependency to it.
2. Pin the publish workflow's gestaltd (`GESTALTD_PINNED_SHA`) to a commit including this gestaltd change.

cc @hughhan1 — this completes the `GESTALT_TARGET_LIBC` scaffolding from the plan-17 build-output work.

### DIFF
--- a/gestaltd/services/apps/providerpkg/prepared_stage.go
+++ b/gestaltd/services/apps/providerpkg/prepared_stage.go
@@ -69,7 +69,9 @@ func StageSourcePreparedInstallDir(manifestPath, stagingDir string, opts StageSo
 		return nil, err
 	}
 	targetOpts := SourceBuildOptions{GOOS: opts.GOOS, GOARCH: opts.GOARCH, Output: opts.BuildOutput}
-	hostBuiltForCatalog, err := ensureHostBuildForSourceStaticCatalog(manifestPath, manifest, SourceBuildOptions{Output: opts.BuildOutput})
+	// Build the catalog binary for the build host's libc so it can be exec'd here
+	// to generate the static catalog (a musl binary cannot run on a glibc runner).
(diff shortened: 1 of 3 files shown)


mzywang:

- #2577 ◀

PR #2581, opened 2026-06-30
Expose custom app operations regardless of allowedOperations

hughhan1:
## Problem

A hybrid app provider that defines custom `@app.operation` handlers *and* a spec (OpenAPI/GraphQL) surface loses those handlers from its effective catalog whenever it also declares `allowedOperations`. The allowlist was applied to the merged plugin + API catalog, so any custom operation not listed in `allowedOperations` was filtered out — at both publish (`provider-release.yaml`) and runtime — and surfaced as a 404 on invocation.

This is what broke `messages.send` on the gmail app (reported in prod): its `allowedOperations` lists only the 12 OpenAPI ops, so the 6 custom write ops (`messages.send`, `messages.reply`, `messages.forward`, `drafts.create/update/send`) were dropped. `hex` has the same latent issue.

## Fix

`allowedOperations` curates the external API surface, not the provider's own hand-written operations. In `resolveExecutableCatalog`:

- The plugin (custom-op) catalog now exposes **every** operation, still honoring alias/description overrides for any listed in `allowedOperations`.
- The API catalog stays gated by `allowedOperations`, and is deduplicated against the plugin catalog so a custom handler wins over a same-named API op (no new merge conflicts).
- Pure-executable providers (no spec surface) are unchanged.

## Test

`TestProviderReleaseMetadataExposesCustomOpsAbsentFromAllowedOperations` exercises the real publish path: a hybrid provider whose `allowedOperations` lists only an API op plus one shared plugin op, with a custom op that is not listed. It fails on `main` (custom op dropped) and passes with this change.

## Rollout

The catalog half of the gmail 404. Once merged, the affected provider snapshots (gmail, hex) must be republished with the new gestaltd so `provider-release.yaml` includes the custom ops. End-to-end `messages.send` also depends on the in-flight musl snapshot fix so the provider binary loads on the Alpine runtime.

### DIFF
--- a/gestaltd/internal/daemon/provider_release_finalize_test.go
+++ b/gestaltd/internal/daemon/provider_release_finalize_test.go
@@ -307,6 +307,68 @@ paths:
 	}
 }
 
+func TestProviderReleaseMetadataExposesCustomOpsAbsentFromAllowedOperations(t *testing.T) {
+	t.Parallel()
+
+	metadata, err := buildProviderReleaseMetadataForRawManifest(t, `kind: app
+source: github.com/testowner/apps/catalog/custom-ops-test
+version: 1.0.0
+displayName: Custom Ops App
+artifacts:
+  - os: test
+    arch: test
+    path: provider
+    sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+entrypoint:
+  artifactPath: provider
+spec:
+  surfaces:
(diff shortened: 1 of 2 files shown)


hughhan1:
Closing: allowedOperations is intended to gate custom @app.operation ops too. The correct fix is to list the missing custom ops in each affected provider's allowedOperations (gmail, hex, etc.), not to change the framework.

PR #2595, opened 2026-07-01
[gestaltd] Add /activate endpoint to trigger deferred app provider startup

mzywang:
## Description

Adds a `POST /activate` endpoint on the management listener that triggers app provider startup via a new `Result.ActivateAppProviders`. Bootstrap still starts providers automatically today, so this is purely additive — the endpoint wraps `providerBuilds.Start` in a `sync.Once` so it's safe to call even after bootstrap already started providers, laying the groundwork for deferring provider startup in a later change.


mzywang:

- #2595 ◀

PR #2598, opened 2026-07-01
[ci] build and publish gestaltd platform binaries concurrently

mzywang:
## Description
The `Build, package, and upload binaries by commit SHA` step in `cd.yml`'s `publish-binaries` job built each of the four target platforms (macos-arm64, macos-x86_64, linux-x86_64, linux-arm64) sequentially in a loop, with each `go build` taking ~80s — about 5m of the job's ~6m runtime. Each platform's build is fully independent (its own build directory, its own GCS object), so this backgrounds all four and waits for them, cutting the build portion from ~4x sequential builds down to roughly one. Per-platform output is buffered to a log file and printed after each finishes so the interleaved concurrent output stays readable, and a failure in one platform no longer aborts the others mid-flight — all platforms still attempt to complete, and the job fails afterward if any did.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -205,22 +205,24 @@ jobs:
             "linux/arm64/linux-arm64"
           )
 
-          mkdir -p build dist
-          for entry in "${platforms[@]}"; do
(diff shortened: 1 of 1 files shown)


mzywang:

- #2598 ◀

PR #2599, opened 2026-07-01
[ci] give the cross-compiled binary build its own Go cache key

mzywang:
## Description
The `publish-binaries` job's `go build` calls (~80s each, for four GOOS/GOARCH targets) never benefited from Go's build cache: `setup-go`'s automatic cache is keyed only by OS + `go.sum`, the same key the lint/test jobs populate with a default `CGO_ENABLED=1` native-arch build. Since this job cross-compiles with `CGO_ENABLED=0` — a flag that changes the build cache ID for nearly every package — none of its compiled objects match that cache, but because a cache already exists under that shared key, this job could never save its own additions back (GitHub's cache service won't overwrite an existing key). This adds a dedicated `actions/cache` step keyed specifically for this job, so its cross-compiled build cache actually persists and gets reused across runs.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -181,14 +181,6 @@ jobs:
         with:
           go-version-file: gestaltd/go.mod
           cache-dependency-path: gestaltd/go.sum
-      # setup-go's own cache is keyed only by OS + go.sum, the same key the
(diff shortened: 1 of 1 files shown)


mzywang:

- #2599 ◀

PR #2600, opened 2026-07-01
[gestaltd] Gate /ready on workflow provider startup

mzywang:
## Description

Gates `/ready` on workflow provider startup by moving `close(workflowProvidersReady)` to *after* `result.StartWorkflowProviders(ctx)` returns successfully (previously it closed before).

The Temporal workflow provider now promotes this build to the deployment's current version inside the `StartProvider` RPC that `StartWorkflowProviders` drives (see gestalt-providers#1106), so a successful return means promotion is confirmed. If promotion fails, the channel never closes, `/ready` never returns, and the Cloud Run deploy times out without shifting traffic — so the rollback story needs no explicit Temporal restore.

Part of the gestaltd deferred-app-startup plan (Step 2). Pairs with [gestalt-providers#1106](https://github.com/valon-technologies/gestalt-providers/pull/1106).

### DIFF
--- a/gestaltd/internal/server/runtime.go
+++ b/gestaltd/internal/server/runtime.go
@@ -264,12 +264,6 @@ func serveRuntime(ctx context.Context, cfg *config.Config, connMaps bootstrap.Co
 			workflowErr <- err
 			return
 		}
-		// The workflow provider promotes this build to the deployment's current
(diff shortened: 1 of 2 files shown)


mzywang:

- #2600 ◀

PR #2601, opened 2026-07-03
[gestaltd] Defer only version-changed app providers to /activate

mzywang:
## Description

Step 4 of the gestaltd-deferred-app-startup plan: split app providers at bootstrap into unchanged (start immediately, block readiness) and version-changed (deferred) groups, so only providers actually being updated are held back.

The startup split machinery (`partition` / `newProviderActivation`) already existed; `appStartupCategory` was a stub that returned `NOOP` for every app, so nothing was ever deferred. This PR implements the actual detection:

- **SHA persistence** — new `app_shas` IndexedDB store (keyed by app name) records the artifact SHA of each app's last *successful* install. `readAppSHAs` loads them at bootstrap before partitioning; `writeAppSHA` records them.
- **Detection** — `newAppStartupCategorizer` compares each app's stored SHA against its current-platform artifact SHA (`providerpkg.CurrentPlatformArtifact`). Match → `AppStartupNOOP` (immediate). Mismatch or no stored SHA (first deploy) → `AppStartupUpdate` (deferred).
- **Write-on-success only** — only the deferred group persists its SHA, and only after a successful install, so a failed install leaves the stored SHA unchanged and the next deploy retries it.
- **`server.autoActivate`** — bypasses the split (all immediate). Defaults to `true` when `K_REVISION` is unset (local dev, where path/source providers have no artifact SHA and would otherwise all look changed on every restart) and `false` on Cloud Run.

### Behavior notes
- Bootstrap still calls `activateAppProviders` immediately, so deferred providers load in the **background** (they no longer block readiness) rather than being fully held for `/activate`. Step 5 removes that auto-start.
- **First post-deploy behavior**: on the first Cloud Run deploy after this ships, no SHAs are stored yet, so *every* provider is treated as changed and loads in the background while readiness returns early. The startup proxy queues requests to not-yet-ready providers during that window. Subsequent deploys only defer actually-changed providers. This is the intended trajectory and is safe, but worth knowing for the first rollout.

## Test plan
- Unit tests cover: SHA read/write round-trip, nil-DB safety, categorizer (match/mismatch/missing/autoActivate/path-source), `autoActivate` resolution (explicit vs K_REVISION default), and write-on-successful-install-only.
- [ ] Deploy and confirm unchanged providers are ready immediately while a changed provider loads in the background; confirm its SHA is persisted after install.

### DIFF
--- a/gestaltd/internal/bootstrap/app_shas.go
+++ b/gestaltd/internal/bootstrap/app_shas.go
@@ -1,0 +1,58 @@
+package bootstrap
+
+import (
+	"context"
+	"fmt"
+
+	idb "github.com/valon-technologies/gestalt/sdk/go/indexeddb"
+	"github.com/valon-technologies/gestalt/server/core/indexeddb"
+	"github.com/valon-technologies/gestalt/server/internal/config"
+	"github.com/valon-technologies/gestalt/server/internal/coredata"
+	"github.com/valon-technologies/gestalt/server/services/apps/providerpkg"
+)
+
+// currentAppSHA returns the artifact SHA256 for the current platform from an
+// app's resolved manifest, or "" when unavailable (e.g. path/source providers
+// that carry no pinned artifact).
+func currentAppSHA(entry *config.ProviderEntry) string {
+	if entry == nil || entry.ResolvedManifest == nil {
+		return ""
+	}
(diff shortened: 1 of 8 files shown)


mzywang:

- #2601 ◀

PR #2603, opened 2026-07-03
[gestaltd] Lazily activate deferred providers on first request per instance

mzywang:
## Description

Follow-up to the deferred-app-startup work (#2602). A single `POST /activate` only reaches **one** Cloud Run instance, so with `min-instances>1` the other instances never start their version-changed providers — and instances added later by autoscaling never do at all. Requests routed to an un-activated instance for a changed provider would queue on the startup proxy indefinitely.

Activation must happen **after traffic has shifted** (so a rolled-back candidate never commits changed-provider side effects), which rules out self-activating at startup. The earliest post-shift signal an instance has is **the first request it serves** — so this makes activation lazy and per-instance:

- The startup proxy gains an optional activation trigger. In `await`, if a deferred provider isn't resolved yet, it fires the trigger (idempotent) before waiting — kicking off the deferred group's install *on that instance*.
- The trigger is `Result.ActivateAppProviders`, so it reuses the existing closed-check, `markActivating`, and server-context wiring (the install survives the request returning and is bounded by the 10-minute timeout).
- Only deferred (update-group) proxies get the trigger; unchanged providers are already published at bootstrap and never need it.

The workflow's explicit `/activate` stays as an immediate warm-up of one instance right after the shift; lazy activation covers the remaining instances (and autoscaled ones) as traffic reaches them.

### Trade-off
The *first* request to a given changed provider on each instance waits for that provider to install (it queues until ready); every subsequent request is fast. Given the "activate only after traffic shift" constraint, someone has to absorb the post-shift install cost — this puts it on the first caller rather than leaving the provider permanently unavailable on that instance.

## Test plan
- `TestStartupProviderProxyLazilyActivatesOnAwait`: awaiting an unresolved deferred proxy fires the trigger once and resolves; a second await doesn't re-fire.
- `TestStartupProviderProxyDoesNotActivateWhenAlreadyResolved`: an already-published provider never fires the trigger.
- Full bootstrap suite passes under `-race`; gofmt clean; golangci-lint 0 issues.

### DIFF
--- a/gestaltd/core/errors.go
+++ b/gestaltd/core/errors.go
@@ -9,4 +9,5 @@ var (
 	ErrMCPOnly             = errors.New("this integration is accessible only via MCP")
 	ErrAmbiguousCredential = errors.New("ambiguous external credential")
 	ErrReconnectRequired   = errors.New("external credential reconnect required")
+	ErrProviderActivating  = errors.New("provider is still activating")
 )
--- a/gestaltd/internal/bootstrap/bootstrap.go
+++ b/gestaltd/internal/bootstrap/bootstrap.go
@@ -1342,7 +1342,7 @@ func Bootstrap(ctx context.Context, cfg *config.Config, factories *FactoryRegist
 	closeAudit = false
 	closeWorkflowsOnError = false
 	closeAgentsOnError = false
-	return &Result{
+	result := &Result{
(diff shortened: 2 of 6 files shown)


mzywang:

- #2603 ◀

PR #2615, opened 2026-07-05
[sdk] Declarative app migrations (TypeScript)

mzywang:
## Description
Adds declarative app migrations to the TypeScript SDK. An app declares a `migrations` list on `defineApp` — declarative schema revisions (add/drop store) or imperative data revisions over a restricted, idempotent-only handle — and the SDK auto-runs them to head during `configure`, guarded by an applied-revision ledger, ledger-ahead (rollback) detection that fails closed, and an advisory lock. SDK-only: no proto or gestaltd changes.

Standalone index add/remove on an existing store is deferred (needs a host-service RPC); this PR supports add/drop store, with indexes set at store creation.

### DIFF
--- a/sdk/typescript/src/index.ts
+++ b/sdk/typescript/src/index.ts
@@ -317,6 +317,17 @@ export {
 } from "./providers/indexeddb.ts";
 export type { IndexedDBQuery as WireIndexedDBQuery, KeyValue } from "./indexeddb.ts";
 export {
+  DEFAULT_LEDGER_STORE,
+  runMigrations,
+  type MigrationHandle,
+  type MigrationResult,
+  type MigrationSet,
+  type MigrationStore,
+  type Revision,
+  type SchemaSpec,
+  type StoreSchema,
+} from "./migrate.ts";
+export {
   PresignMethod,
   S3,
   S3ObjectAccess,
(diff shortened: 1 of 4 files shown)


mzywang:

- #2615 ◀

mzywang:
Superseded by #2632 (migration runner) + #2631 (CreateIndex/DeleteIndex RPCs, merged).

Comparison — #2632 is kept because it aligns with the plan's design principles that this PR predates:

- **Lock correctness.** This PR uses a client-side lock (`add` insert-if-absent + stale-check + `put` reclaim). That's exactly the race the plan's "Why the lease is a host-service RPC" section rejects: two instances that both observe an expired lease each `put` and both win, running concurrently. #2632 uses the atomic `AcquireLock`/`ReleaseLock` host-service lease (Step 1, #2619) that was built specifically to avoid this.
- **Four operations.** This PR supports only add/drop store (indexes at creation only). #2632 supports the full four declarative ops — add/drop store and add/remove index on an existing store — via the CreateIndex/DeleteIndex RPCs (#2631).
- **Extras #2632 adds:** lease renewal during long migrations, and a richer restricted handle (read-only index access + transactions).

The one nice touch from here — `runMigrations` returning `{ applied, head }` — has been folded into #2632.

PR #2628, opened 2026-07-06
[sdk] CreateIndex/DeleteIndex on the IndexedDB service

mzywang:
## Description

Adds `CreateIndex`/`DeleteIndex` RPCs to the IndexedDB service and wires them end-to-end through gestaltd's typed host-service proxy, giving the migration runner the two remaining declarative schema operations (add/remove an index on an existing store) alongside create/drop object store. Includes the regenerated bindings (Go server, Go/Python/Rust/TS SDK, wire stubs), gestaltd forwarding, the optional `IndexedDBIndexProvider` capability + serve handlers, and the native TS client `createIndex`/`deleteIndex`.

Index management is optional: each layer returns `Unimplemented` until a backend implements it (relationaldb: valon-technologies/gestalt-providers#1114).

Re-introduces the change reverted in #2627 (originally pushed to `main` by accident); stacked on that revert.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2627, opened 2026-07-06
Revert "[sdk] CreateIndex/DeleteIndex on the IndexedDB service"

mzywang:
## Description

Reverts commit 7a8a2c795, which was pushed directly to `main` by accident (a stale branch upstream made `gs submit` target `main`). This restores `main` to not contain the CreateIndex/DeleteIndex change, which will be re-introduced for review in a follow-up PR. #2626 and all later history are preserved.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2629, opened 2026-07-06
[sdk] Declarative migrations runner for TypeScript apps

mzywang:
## Description

Lets an app author declare an ordered list of revisions on `defineApp({ migrations })` and has the SDK discover and run them during provider startup — the author never calls a runner. Includes the runner algorithm (ledger store, `AcquireLock` lease with renewal + expiry-wait, ledger-ahead-of-code detection that fails closed, ordered pending set, apply-then-record with idempotent replay), a restricted data-only `MigrationHandle` (no `add`, no DDL, no `close`), and the declarative four-operation schema diff (add/drop store, add/remove index via create-if-absent / drop-if-exists).

`AppProvider.configureProvider` opens the DB binding, runs migrations before the author's configure handler, and always releases the lease and closes the connection.

Stacked on #2628 (needs its CreateIndex/DeleteIndex client methods).


mzywang:

- #2627
    - #2628
        - #2629 ◀

mzywang:
Superseded — recreating cleanly stacked on a revert of the accidental main pushes.

PR #2642, opened 2026-07-07
[sdk] Idempotent backfill revisions (runner-driven, from != into)

mzywang:
## Description

Replaces the imperative `up(handle)` backfill escape hatch with a runner-driven `BackfillRevision` that is **idempotent by construction**:

- **`BackfillRevision { from, into, value }`** — the runner streams the immutable source `from`, computes the pure `value(row)`, and upserts into a **distinct** `into`. `from === into` is rejected at validation, so a backfill can never read its own output; read-modify-write (`{ ...row, count: row.count + 1 }`) is inexpressible and there are no in-place data writes.

Removes `ImperativeRevision` and the `MigrationHandle`/`Store`/`Index`/`Transaction` surface. Because non-idempotency is unrepresentable, no CI idempotency gate is needed. Field-level shape evolution stays a read-time concern (JSON payload) and needs no migration.


mzywang:

- #2642 ◀

mzywang:
Thanks Bugbot — addressed both:

**Schema null misroutes backfill** — real bug, fixed. `isSchemaRevision` used `schema !== undefined` while `validateRevisions` used `schema != null`, so `{ schema: null, backfill: {...} }` passed validation but routed to `applySchema(null)`. Aligned `isSchemaRevision` to `schema != null` (latest commit) and added a regression test (`a null schema alongside a backfill runs the backfill, not applySchema`).

**Lease unchecked before backfill writes** — not a regression, no change. In `applyBackfill` there is no `await` between `assertHeld()` and `target.put(...)` (only the synchronous `cursor.value` read and `value(row)` call). The lease-loss flag is set by the renewal `setInterval` callback, which can only run at an `await` point, so it cannot flip in that window — `assertHeld()` guards the write as tightly as the old handle's in-`put` check did. Loss *during* the put's own await is unavoidable and design-tolerated: idempotency is the correctness guarantee, and the lease only prevents redundant concurrent work (a write completing after a lapse is convergent, not corrupting).

PR #2643, opened 2026-07-07
[sdk] Migration idempotency test harness

mzywang:
## Description

Adds `verifyRevisionIdempotent` / `verifyAllRevisions` and a `MemoryIndexedDB` fake, exported from a new `./testing` entrypoint. Every declared revision is double-applied (replay) and applied concurrently against the in-memory DB; if the resulting state diverges from a single application it throws `IdempotencyError`. Apps get a one-line CI gate: `verifyAllRevisions(app.migrations)`.

Also exports `applyRevisionForVerification` (apply one revision without the ledger/lease) so replay and concurrent convergence can be exercised directly.

Implements Step 2 of the idempotent-backfills plan. Stacked on #2642.


mzywang:

- #2642
    - #2643 ◀

mzywang:
Closing: superseded by the `from !== into` enforcement added to #2642. Banning in-place backfills (source must differ from target) makes every backfill idempotent by construction — the runner-driven pure transform reads an immutable source and upserts a distinct target, so it can never read its own output. That removes the residual gap this CI harness existed to catch, so the double-apply/concurrent-apply gate is no longer needed.

PR #2799, opened 2026-07-15
Resolve app registry installs when config app key differs from registry app name

mzywang:
## Summary
- Look up deploy config apps by registry app name when the direct `apps.<name>` config key does not match.
- Match pinned apps via manifest `source` (`apps/{app}`) or `displayName`, so first-time installs can resolve `from_version` from the config pin.

## Problem
Dev gestaltd deploy keys g-issues as `apps.gIssues`, but the registry/admin install API uses the published app id `g-issues`. The installer only checked `ConfigApps[appName]`, so first install failed with:

```
resolve from_version: no known fleet version and app is not pinned in config
```

## Approach
Prefer a gestalt installer fix over renaming deploy config keys, since config naming conventions may differ from published registry app ids across environments.

## Test plan
- [x] `go test ./gestaltd/internal/appregistry/... -count=1`
- [ ] After gestaltd deploy: `POST /admin/api/v1/app-registries/toolshed/apps/g-issues/install` succeeds on dev

## Follow-up
After merge, bump `GESTALTD_COMMIT_SHA` in valon-tools and deploy. Pair with valon-technologies/valon-tools#222 for admin API auth on dev/stage.

### DIFF
--- a/gestaltd/internal/appregistry/config_app.go
+++ b/gestaltd/internal/appregistry/config_app.go
@@ -1,0 +1,31 @@
+package appregistry
+
+import (
+	"strings"
+
+	"github.com/valon-technologies/gestalt/server/internal/config"
+)
+
+func resolveConfigAppEntry(configApps map[string]*config.ProviderEntry, registryAppName string) *config.ProviderEntry {
+	registryAppName = strings.TrimSpace(registryAppName)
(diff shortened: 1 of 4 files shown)


mzywang:
Closing in favor of valon-tools deploy config rename (apps.gIssues → apps.g-issues).

PR #2842, opened 2026-07-19
Consolidate identity/authorization onto providergateway

hughhan1:
## What

Collapses six hand-rolled `CheckAccess` request-building / subject-derivation sites onto two shared helpers in `services/invocation`, and deletes the duplicated legacy `/api/v1/authorization/*` REST surface. Enforcement outcomes are byte-for-byte identical; the gateway's shadow-mode `Authorize` path stays shadow.

## Stacked on

#2840 (CLI authorization subcommands → `/api/v2`). No file overlap; no CLI work in this branch.

## New shared surface (`services/invocation/authorization_check.go`)

```go
func SubjectAccessRequest(subjectID, action string, resource *proto.Resource) *proto.CheckAccessRequest
func CheckSubjectAccess(ctx context.Context, authorization core.AuthorizationProvider, req *proto.CheckAccessRequest) (bool, error)
```

`CheckSubjectAccess` maps nil-response / error → `(false, err)`, else `resp.GetAllowed()`. No new exported surface on `providergateway`; `core.IdentityProvider` / `core.AuthorizationProvider` and all existing gateway signatures untouched.

## Deleted API endpoints

```diff
- POST   /api/v1/authorization/check-access
- GET    /api/v1/authorization/relationships
- GET    /api/v1/authorization/models/active
- GET    /api/v1/authorization/models/active/resource-types
```

The four routes, `handlers_authorization.go` (4 handlers + protojson/query helpers), and `authorization_api_test.go` are deleted. `requireAuthorizationProvider` / `authorizationHTTPContext` (the only survivors, used by workflow-run REST) relocate to `handlers_workflow_runs.go`. The `/api/v1/authorization` reserved-namespace entry is dropped from `http_bindings.go`.

## Request lifecycle: before

```mermaid
sequenceDiagram
    autonumber
    participant C as Caller
    participant S as internal/server
    participant B as invocation.Broker
    participant WM as workflows.ManagerServer
    participant GW as providergateway
    participant P as AuthorizationProvider

    Note over C,P: Six sites each hand-roll CheckAccess
    C->>S: POST /api/v1/authorization/check-access
    S->>P: CheckAccess (hand-built req)
    C->>S: workflow run / service-account mgmt
    S->>P: CheckAccess (hand-built req, local subjectID)
    C->>B: app invoke
    B->>B: accessRequest() builds req + Subject.Properties
    B->>P: CheckAccess
    C->>WM: workflow RPC
[...]

PR #2848, opened 2026-07-19
bootstrap: wire workflow host service to the Raw authorization provider (ProviderGateway rollout Step 1)

hughhan1:
## Summary

Step 1 of the ProviderGateway enforcement rollout. Splits the bootstrap authorization wiring so the plugin-facing workflow `ProviderServer` uses the direct (Raw) authorization provider for gestaltd's own enforcement checks, while the authorization host service keeps the gateway-guarded (Guarded) client.

## Context

- #2842 collapsed the six hand-rolled `CheckAccess` sites onto shared helpers and deleted `/api/v1/authorization/*`.
- #2840 moved the CLI to `/api/v2`.
- #2820 (Jul 17) flipped `ProviderGatewayTransport.Authorize` from `return true, nil` (true shadow) to `return allowed, nil`. **Enforcement is already ON in code on main**, and it replaced the CallerToken subject mechanism with `principal.FromContext` plus the capability relay.
- Production valon.tools still runs a pre-#2820 gestaltd image, so prod behavior is still shadow. Datadog (`gestaltd.provider_gateway.authorization.count`, 30d): **26.9M `gd.allowed:false` vs 39 `gd.allowed:true`**.
- The Guarded (gateway-wrapped) authorization client backs two things: the **authorization host service** (plugins calling Authorization RPCs; the intended security boundary) and the **plugin-facing workflow `ProviderServer`** (`host_service_bootstrap.go:477`). The latter is a wiring bug: `requireWorkflowAccess` / `requireServiceAccountManagement` are gestaltd's *own* enforcement checks, and meta-checking them violates the non-recursion principle. The public-surface workflow server already correctly uses the Raw provider.

Net without this fix: the next routine gestaltd release flips hard enforcement with a near-100% deny rate on the guarded paths, including gestaltd's own workflow enforcement checks.

## Changes (all in `gestaltd/internal/bootstrap`)

- `bootstrap.go`: add `Deps.AuthorizationInternal` (`core.AuthorizationProvider`) next to `Deps.Authorization`, with a comment documenting the split — `Authorization` is the gateway-guarded (Guarded) client for plugin-facing host services; `AuthorizationInternal` is the direct (Raw) client for gestaltd's own enforcement checks.
- `bootstrap.go` (`prepareCore`): populate `Deps.AuthorizationInternal` via the existing `selectedAuthorizationProviderInstance(cfg, authorizationProviders.Raw)` helper, mirroring the Guarded selection.
- `host_service_bootstrap.go`: `buildWorkflowProviderHostService` passes `deps.AuthorizationInternal` into `workflowservice.NewProviderServer` instead of `deps.Authorization`. `authorizationHostServiceFromDeps` stays on `deps.Authorization` (Guarded); that meta-check is the intended plugin boundary.

No provider-gateway surface changes. No behavior change on shadow-era builds; on post-#2820 builds this removes only the meta-check in front of the workflow server's own `CheckAccess` calls (the real workflow authorization check is unchanged). Nil stays nil: if no authorization provider is configured, the workflow server already errors `FailedPrecondition`.

## Tests

- Added `TestBuildProviderHostServicesWiresAuthorizationSplit`: a bootstrap-level assertion that the authorization host service consults the Guarded provider and the workflow host service consults the Raw provider, using sentinel recording providers and in-memory gRPC RPCs (`CheckAccess` for authorization, `ListRuns` for workflow, which runs `requireWorkflowAccess` → `CheckAccess` before the manager call). Verified the test **fails** when the workflow host service is wired back to the Guarded provider.
- Updated five `executable_app_test.go` `Deps` literals that built hosted apps with a single allow-all provider to also set `AuthorizationInternal`, so the workflow host service keeps receiving an allow-all provider. These pinned the old single-provider wiring.

## Verification

```
go build ./...
go test ./internal/bootstrap/... ./services/workflows/... ./services/providergateway/... ./internal/server/...
go vet ./...
```

All green from `gestaltd/`.

## Release caveat

⚠️ **This fix alone does NOT make a gestaltd release safe.** Plugin callbacks to the authorization host service still hard-enforce on post-#2820 builds, and the authorization model grants almost nobody access to the authorization provider resource (`provider/indexeddb` in prod). **Hold gestaltd releases to valon-tools until Step 2/3 are decided** (meta-check policy + seed grants, then stage bake with fresh signal).

## Out of scope (documented for planning)

Steps 2-5 of the rollout (meta-check policy/seed grants, stage bake, prod release, post-enforcement cleanup) are intentionally not implemented here. This PR is Step 1 only.

### DIFF
--- a/gestaltd/internal/bootstrap/bootstrap.go
+++ b/gestaltd/internal/bootstrap/bootstrap.go
@@ -160,21 +160,28 @@ func (m providerMetadata) descriptionOr(v string) string {
 type Deps struct {
 	// EncryptionKey is the derived 32-byte key from server.encryptionKey, not the
 	// raw config value.
-	EncryptionKey           []byte
-	BaseURL                 string
-	RuntimeRelayBaseURL     string
-	SecretManager           core.SecretManager
-	Services                *coredata.Services
-	SelectedIndexedDBName   string
-	IndexedDBs              map[string]indexeddb.IndexedDB
-	IndexedDBDefs           map[string]*config.ProviderEntry
-	IndexedDBFactory        IndexedDBFactory
-	Caches                  map[string]corecache.Cache
-	CacheDefs               map[string]*config.ProviderEntry
-	CacheFactory            CacheFactory
-	S3                      map[string]s3sdk.S3
-	Authentication          core.IdentityProvider
-	Authorization           core.AuthorizationProvider
+	EncryptionKey         []byte
+	BaseURL               string
+	RuntimeRelayBaseURL   string
+	SecretManager         core.SecretManager
+	Services              *coredata.Services
+	SelectedIndexedDBName string
+	IndexedDBs            map[string]indexeddb.IndexedDB
+	IndexedDBDefs         map[string]*config.ProviderEntry
+	IndexedDBFactory      IndexedDBFactory
+	Caches                map[string]corecache.Cache
+	CacheDefs             map[string]*config.ProviderEntry
+	CacheFactory          CacheFactory
+	S3                    map[string]s3sdk.S3
+	Authentication        core.IdentityProvider
+	Authorization         core.AuthorizationProvider
+	// AuthorizationInternal is the direct (Raw) authorization provider used for
+	// gestaltd's own enforcement checks (e.g. the workflow ProviderServer's
+	// requireWorkflowAccess/requireServiceAccountManagement), which must not be
+	// meta-checked by the provider gateway. Authorization above remains the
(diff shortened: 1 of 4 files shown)


hughhan1:
Closing this PR. The Guarded/Raw authorization provider split was collapsed in #2858 (merged to main). This PR introduces `AuthorizationInternal` to split Guarded/Raw wiring for the workflow host service, but the Guarded path no longer exists — `authorizationProviderSets` now has only `Raw`. The meta-check recursion problem this PR was solving is gone.

PR #2870, opened 2026-07-20
App registry step 12: implement registry-only app lifecycle

mzywang:
## Summary
- add validated `source.registry` app config and registry-only lock/sync entries without baked artifacts
- split registry admission into `add` and `upgrade` flows with fleet-catalog `from_version` semantics
- bootstrap registry-only providers from the latest known version, sharing materialization with the poller and activating version-synchronized static mounts

## Stack
- Depends on #2868 (`app-registry-step12-docs`)

## Testing
- `cd gestaltd && go test ./...`


mzywang:
Superseded by #2872, which is the cleaner implementation and now includes runtime registry static-mount support.

mzywang:
Reopening this implementation as the retained step 12 PR; #2872 will be closed instead.

PR #2892, opened 2026-07-22
Document workflow target validation for app registry install admission

mzywang:
## Summary
- Extend `plans/app_registry/validation.md` with install-time checks for packaged workflow app targets and operations.
- Document what gestaltd should inspect (`steps[].app.name` against deploy `config.apps`, operations against target `interface`) and the current bootstrap-only enforcement gap.
- Add the `g-issues` / `gIssues` rename failure as a concrete example.

## Test plan
- [x] Docs-only change; no code or tests required.


mzywang:
Punting on install-validation spec for now; keeping plan.md future-work documentation in #2895.

PR #2893, opened 2026-07-22
Validate packaged workflow targets at app registry install time

mzywang:
## Summary
- Reject registry installs when packaged workflow app-call steps reference app slots missing from deploy `config.apps`, or when fleet-known targets do not publish the referenced operation.
- Copy workflow app-call metadata from `workflows.yaml` into provider release static validation and registry `versions/{version}.json` entries at publish time.
- Generate `workflows.yaml` during provider packaging via `GESTALT_APP_WRITE_WORKFLOWS` (Go and TypeScript SDKs), alongside existing static catalog generation.

## Test plan
- [x] `go test ./internal/appregistry/...`
- [x] `go test ./internal/providerrelease/...`
- [x] `go test ./services/apps/providerpkg/...`
- [x] `go test ./internal/daemon -run ProviderRelease`
- [ ] After merge and republish, installing a snapshot whose workflow still targets `gIssues` should fail with `workflow_target_app_missing` before IndexedDB admission

Stacked on #2892 (docs).


mzywang:
Punting on implementation for now; keeping plan.md future-work documentation in #2895.

PR #2906, opened 2026-07-22
Fix admin registry summary version overflow

mzywang:
## Summary
- Prevent long version strings from overflowing the Summary card on `/admin/registry/$app`.
- Use a 2-column summary grid with `min-w-0` field wrappers.
- Teach `Code` to wrap (`break-all`, `max-w-full`) for digest-style values.

## Test plan
- [ ] Open `/admin/registry/<app>` with a long desired/published version and confirm Summary stays within the card


mzywang:
Addressed in the latest commit: cohort totals stay as the primary view, and a collapsible **Cohort replicas** table restores instance-level restart/attempt/error detail for replicas in the current enrollment cohort (`inCohort` only). This keeps the detail page lighter than the old full materializations table while preserving rollout debugging.

PR #2993, opened 2026-07-29
helm: default to RollingUpdate and stabilize Service ClusterIP

hughhan1:
## Summary
- Switch the gestaltd chart default rollout strategy from `Recreate` to `RollingUpdate` with `maxUnavailable: 0`, `maxSurge: 1`, so old pods keep serving until new ones pass readiness. The previous `Recreate` default killed all pods before starting new ones, leaving the `gestaltd` ClusterIP Service with zero ready endpoints on every deploy and causing the Cloudflare tunnel in front of it to return 502 for ~2 minutes.
- Pin the single-replica local profile (`values-local.yaml`) to `Recreate` explicitly, since its `ReadWriteOnce` PVC makes a `RollingUpdate` surge pod wait on a volume the old pod still holds.
- Expose `service.clusterIP` and `service.publishNotReadyAddresses` in the Service template and values. Pinning `clusterIP` keeps the address stable across reinstalls so clients that cache the resolved ClusterIP (e.g. `cloudflared`) don't keep dialing a dead old IP after the Service is recreated. `publishNotReadyAddresses` defaults to `false`, pairing with the `RollingUpdate` default so traffic only routes to ready pods.

## Test plan
- [x] `helm lint gestaltd/deploy/helm/gestaltd` passes
- [x] `helm template` renders `RollingUpdate` / `publishNotReadyAddresses: false` under base, `values-local`, `ci-values`, and `ci-values-preparation`
- [x] `values-local.yaml` renders `Recreate` (preserved behavior)
- [x] `service.clusterIP` is omitted when empty, rendered when set
- [ ] CI `validate-gestaltd` (`helm lint` + `helm template` matrix) passes

### DIFF
--- a/gestaltd/deploy/helm/gestaltd/templates/service.yaml
+++ b/gestaltd/deploy/helm/gestaltd/templates/service.yaml
@@ -6,6 +6,10 @@ metadata:
     {{- include "gestaltd.labels" . | nindent 4 }}
 spec:
   type: {{ .Values.service.type }}
+  {{- with .Values.service.clusterIP }}
+  clusterIP: {{ . }}
+  {{- end }}
+  publishNotReadyAddresses: {{ .Values.service.publishNotReadyAddresses | default false }}
   ports:
     - port: {{ .Values.service.port }}
       targetPort: http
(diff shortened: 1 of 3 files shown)


hughhan1:
Closing; pivoting to a simpler approach.

PR #3046, opened 2026-08-04
gestaltd: advertise workflowDefaultProvider on auth/info

gvocale:
## Summary

`GET /api/v1/auth/info` advertises `features.workflowDefaultProvider` when a default workflow platform provider is selected, so console Workflows can call the provider-scoped Workflow API without guessing engine names.

## Why / context

App admin Workflows resolves the Workflow API `provider` query from auth/info. Without this field, Valon showed “Workflows are not configured” even though `provider=local` already served runs. Hardcoding or probing provider names in the console would encode deploy internals in the frontend; the deployment owns the engine identity and should publish it on the existing capabilities surface (same pattern as `features.agent`).

## What changed

- Auth/info features include `workflowDefaultProvider` from `WorkflowControl.DefaultProviderName()`.
- The field is omitted when no default is selected (empty name), including when `WorkflowControl` is present.
- Workflow event delivery uses the same `WorkflowControl` read path (no duplicated `workflowProviderName` snapshot).
- HTTP API docs note the new feature field.

## Validation

- [x] Automated: `go test ./internal/server/ -run 'TestAuthInfo|TestWorkflowEvent'`
- [ ] Manual: after deploy, `GET /api/v1/auth/info` includes `features.workflowDefaultProvider` matching the deploy’s workflow provider (e.g. `local` on valon.tools)
- [ ] Manual: console `/apps/<app>/admin/workflows` loads without the configuration banner once the field is advertised

## Reviewer focus

Confirm `WorkflowControl` as the single owner of engine identity, and empty `DefaultProviderName()` as the “not configured” gate (not nil control).

## Rollout / compatibility

Additive JSON field with `omitempty`. Older clients ignore it; newer console clients keep failing closed until the field appears. No config or migration change required beyond shipping gestaltd.


mzywang (approved):
Low-risk additive capability discovery change. Existing workflow execution remains unchanged, and the provider source-of-truth consolidation looks safe.

PR #3070, opened 2026-08-07
[gestaltd] canonicalize identity at every authorization boundary

mzywang:
## Description

Resolves verified subjects to the persisted `user:<uuid>` before the mounted-UI and app-admin boundaries authorize, and fails closed instead of falling back to the raw token subject.

Root cause addressed: `resolvePrincipalUserID` treated any subject without an `@` as already canonical, so a provider-opaque subject such as `auth0|abc` reached authorization unresolved and matched no relationship — the mechanism behind the employee access loss. Subjects are now classified explicitly (canonical UUID short-circuits with no user-store round trip, verified email resolves, opaque is rejected).

Stack A / G2 of [plans/external_users/implementation_v2.md](https://github.com/valon-technologies/gestalt/blob/main/plans/external_users/implementation_v2.md). Stacked on #3069. Relationship traversal is unchanged here — unifying the UI/catalog/admin scans into one evaluator is G3.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3071, opened 2026-08-07
[gestaltd] decide mounted UI and app admin through the evaluator

mzywang:
## Description

Replaces the bespoke direct-only `ListRelationships` scans in the mounted-UI and app-admin boundaries with one shared `CheckAccess`-based decision helper, so group and subject-set grants are honored where previously only direct grants were visible. Also centralizes app key → resource type / policy alias mapping in `invocation.AuthorizationResourceMapper`, shared by the broker and the server.

Stack A / first half of G3 in [plans/external_users/implementation_v2.md](https://github.com/valon-technologies/gestalt/blob/main/plans/external_users/implementation_v2.md). Stacked on #3070.

### Transition shim — please review this specifically

`CheckAccessRequest` is strictly action-shaped and there is no effective-roles RPC, so a resource type that declares neither a matching action nor a `"*"` wildcard makes the evaluator answer "denied" to a question it cannot represent. Treating that as a real denial would lock existing direct-grant holders out of their mounted UI — including the built-in admin UI, whose policy typically declares no actions.

`legacyDirectGrantMountedAccess` detects that case *from the successfully-read active model* and keeps the legacy direct-grant path working, with a `WarnContext` naming the resource type and action. It is unreachable from any provider or transport error path (those always deny), it can only allow (never widen a denial into an allow beyond the legacy behavior), and its scan is bounded at 20 pages where the original loop was unbounded.

The warning is the signal for T1: if it never fires in production, the shim is safe to delete.

### Known follow-ups (not addressed here)
- `hasExplicitAppAdmin` is deliberately **not** shimmed — it asks about the shared `app` resource type using the app key as the action, which `CheckProviderAccess` already exercises in production against a wildcard action. If T1 finds a registry app with a dedicated resource type declaring no matching action, that path needs the same treatment.
- AllowedRoles gating and app admin now depend on providers populating `matched_relations`. The broker has depended on this since #3068, but G4's conformance suite should assert it against the real evaluator.
- App-admin resource mapping now honors a configured `authorizationPolicy` alias instead of hardcoding `{app, <name>}` — a live behavior change for any registry app that sets one. Worth a config sweep before deploy.


mzywang:

PR #3069, opened 2026-08-07
[gestaltd] gate authorization state startup writes behind explicit apply

mzywang:
## Description

Gates `bootstrapAuthorizationProviderState` behind an explicit apply switch (`Server.AuthorizationStateApply`, falling back to `GESTALTD_AUTHORIZATION_STATE_APPLY`) so a no-traffic candidate no longer overwrites shared authorization state just by starting up. Default is now plan-only: it computes the static model and logs its digest without calling `SetAuthorizationState`.

Stack A / G1 of [plans/external_users/implementation_v2.md](https://github.com/valon-technologies/gestalt/blob/main/plans/external_users/implementation_v2.md), addressing incident finding #5. Snapshot storage, rollback command generation, and a CLI plan/apply/rollback surface are out of scope here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3072, opened 2026-08-07
[gestaltd] batch catalog, operation and MCP listing decisions

mzywang:
## Description

`FilterCatalogForPrincipal` ignored its principal and returned the catalog unfiltered, and MCP `tools/list` filtered only on token scope — so listings did not reflect authorization and enforcement happened only at invoke time. Both now decide through the same evaluator path used for invocation, batched via `CheckAccessMany`.

Stack A / second half of G3 in [plans/external_users/implementation_v2.md](https://github.com/valon-technologies/gestalt/blob/main/plans/external_users/implementation_v2.md). Stacked on #3071.

### Why this doesn't cause access loss

Listing filtering is a new restriction, so semantic identity with the single-decision path matters more than the batching itself. Batching is implemented as a request-scoped **prefetch into a cache keyed on the exact `(subject, action, resourceType, resourceID, allowedRoles)` question** — not as a second decision path. `/apps` decision code is unchanged; the `defaultRole` fallback and the G3a transition shim still run per item.

- A failed, truncated, or malformed batch logs and returns an empty cache, so every item falls back to the per-item decision. Listing degrades to more calls, never to fewer results.
- Transport errors surface as 503 (`/apps`, `/apps/{name}/operations`) or a JSON-RPC error (`tools/list`) rather than an empty list that reads as "you have no apps".
- Regression guard included: a subject visible only via `defaultRole` still sees its apps.

Call-count evidence in tests: `/apps` over two apps → 1 `CheckAccessMany`, 0 per-item `CheckAccess`; same for catalog, operation listing, and `tools/list`.

### Behavior changes worth reviewing
- **User lookup is now gated on an explicit operator role** (`server.userLookup`, defaulting to resource `gestaltUserLookup` / relation `operator` whenever auth and an authorization provider are configured). App admins still see the full grant roster for their own app — only subject-ID→email resolution is withheld, so admin surfaces can't be used as a directory. **This means app admins lose email visibility until granted the operator role.** Empty policy (no-auth deployments) leaves behavior unchanged.
- `/apps/{name}/operations` now also applies the token-scope filter. Browser sessions have empty scopes and are unaffected; narrow API tokens see a list matching what they can actually call.
- `tools/list` with a nil principal now lists nothing (previously everything). `tools/call` already refused nil principals, so this closes a listing/enforcement gap.
- Catalog/MCP operation filtering has **no** transition shim — it mirrors `authorizeOperation` exactly, so nothing invocable can be hidden, but an app resource type declaring neither the operation ID nor `"*"` will see those operations disappear from listings. They already fail at invoke, so this is truthful rather than new access loss. T1's inventory should still confirm every app resource type declares `"*"` before Gate A.

### DIFF
--- a/gestaltd/internal/config/config.go
+++ b/gestaltd/internal/config/config.go
@@ -1938,6 +1938,17 @@ type AdminConfig struct {
 	AllowedRoles        []string `yaml:"allowedRoles,omitempty"`
 }
 
+// UserLookupConfig gates resolving other people's identities (turning a subject
+// ID into an email on an admin surface) on an explicit employee operator role.
+// It is deliberately separate from any app's admin policy: app-scoped
+// administration alone must not permit user enumeration. Leaving it unset uses
+// the built-in operator resource and relation whenever authorization is
+// configured.
+type UserLookupConfig struct {
+	AuthorizationPolicy string   `yaml:"authorizationPolicy,omitempty"`
+	AllowedRoles        []string `yaml:"allowedRoles,omitempty"`
+}
+
 type ServerConfig struct {
 	Public        ListenerConfig           `yaml:"public"`
 	Management    ManagementListenerConfig `yaml:"management"`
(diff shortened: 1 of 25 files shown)


mzywang:

mzywang:
Moved into Phase 1 of the external-users rollout. The Phase 1 conformance suite (#3073) covers catalog and MCP `tools/list` decisions, and those surfaces only authorize once this lands — #3073 does not compile against #3071 alone, which is why it was stacked here originally.

Landing this in Phase 1 is code-only reordering: `defaultRole: viewer` on `app` is still active, so every authenticated employee resolves viewer and the listing filters nothing. Deployment sequencing stays controlled by the `GESTALTD_PINNED_SHA` bump in Toolshed. The Phase 4 prerequisite — verifying an ungranted identity sees an empty `/apps` and an empty `tools/list` — is unchanged and still gates enabling any external domain.

Merge order: #3069 → #3070 → #3071 → #3072 → #3073 → [toolshed#4144](https://github.com/valon-technologies/toolshed/pull/4144).

PR #3073, opened 2026-08-08
[gestaltd] authorization conformance suite across every server surface

mzywang:
## Description

Adds a conformance suite driving mounted UI, catalog/`/apps`, app admin, user lookup, operation invocation, and MCP `tools/list`/`tools/call` against one shared evaluator — proving every surface asks the same question and projects the answer the same way. This is the check that would have caught the incident: #3061 broke group-only catalog access precisely because each surface traversed relationships independently.

Stack A / G4 in [plans/external_users/implementation_v2.md](https://github.com/valon-technologies/gestalt/blob/main/plans/external_users/implementation_v2.md). Stacked on #3072.

### ⚠️ Read this before approving: it is NOT the real evaluator

G4 asks for "the real authorization evaluator". **That is not achievable from this module, so this suite uses an in-repo reference implementation instead.** Being explicit because the plan's whole point is not trusting untested assumptions.

Why (a) is impossible: the real evaluator (`gestalt-providers/authorization/indexeddb/access.go`) is a **separate Go module in another repository**, its own `go.mod` `replace`s back into this checkout via relative paths, and it requires a live indexeddb host service. CI runs `go test` inside the `gestaltd` module with no sibling checkout and no network. There is no in-repo authorization provider — `buildNamedAuthorizationProvider` only builds a remote gRPC client or calls a factory nothing in-repo registers.

`conformanceEvaluator` was transcribed line-by-line from the real `evaluateAccess` (resource type absent → deny; action lookup then `"*"` fallback; empty relation set → deny; relation must be in the action's allowed set; recursive subject-set expansion with a visited set; matched relations in declared order; `defaultRole` only when the action permits it). It is the **single** evaluator every case uses, reached through one constructor, and the suite is written against `core.AuthorizationProvider` without touching internals — so swapping in the real provider is a one-function change.

**Proven:** every server surface agrees with every other. **Not proven:** that the shipped provider implements these semantics. That remains Gate A's canary and T2's shadow evaluation.

### Coverage
All G4 bullets are covered: direct / one-level / nested / cyclic / malformed / paginated grants; group-derived admin; policy aliases and dedicated resource types; employee, elevated admin, ungranted external, granted external, revoked external; external app-admin isolation and user-lookup denial; browser, CLI token exchange, API token, and MCP list + call. Cyclic and malformed assert fail-closed *and bounded* (cycle detection, not a budget brake, terminates the walk).

Includes the explicit assertion for the risk flagged in #3071: a provider returning `allowed: true` with empty `matched_relations` must not silently grant a role-gated mount or app admin.

The assertions were mutation-tested — removing the broker's `OperationAccess`, the authorization provider, or the `AuthorizationPolicies` map each makes specific tests fail rather than silently pass.

Suite cost is ~0.1s inside a 0.76s package run, and it is `-race` clean.

### Note on an unrelated flaky test
`TestAgentRuntimeConfigSelectedProviderStartsSessionWithRuntimeFields` (`internal/bootstrap`) intermittently times out at ~5s under full-suite parallel load. I confirmed it fails identically on unmodified `origin/main`, so it is pre-existing and unrelated to this stack.

### DIFF
--- a/gestaltd/internal/server/conformance_evaluator_test.go
+++ b/gestaltd/internal/server/conformance_evaluator_test.go
@@ -1,0 +1,516 @@
+package server_test
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/valon-technologies/gestalt/server/core"
+	proto "github.com/valon-technologies/gestalt/server/rpc/protov1/v1"
+)
+
+// This file holds the single authorization evaluator the conformance suite in
+// conformance_test.go runs every case against.
+//
+// HONEST PROVENANCE. The production relationship-graph evaluator lives outside
+// this repository, in the gestalt-providers module
+// github.com/valon-technologies/gestalt-providers/authorization/indexeddb. It
+// is a separate Go module that is not in this repository's go.work, depends on
+// the indexeddb host service at runtime, and would introduce a module cycle
+// (it `replace`s the gestalt sdk and rpc modules back onto this checkout). CI
+// runs `go test ./...` in the gestaltd module with no sibling checkout and no
+// network, so that package cannot be built or run here.
+//
+// conformanceEvaluator is therefore a reference implementation of the documented
+// evaluation semantics, transcribed from that provider's evaluateAccess:
+//
+//   - a resource type absent from the active model denies;
+//   - an action is looked up by name, then by the "*" wildcard action;
+//   - an action with no allowed relations denies;
+//   - a relationship counts only when its relation is in the action's allowed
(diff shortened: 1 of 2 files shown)


mzywang:

mzywang:
Top of the Phase 1 stack. Merge order: #3069 → #3070 → #3071 → #3072 → #3073, then [toolshed#4144](https://github.com/valon-technologies/toolshed/pull/4144) once `GESTALTD_PINNED_SHA` is bumped past this stack.

Base is #3072 rather than #3071 because the suite asserts catalog and MCP listing behavior that only exists there; it does not build against #3071 alone.

PR #3115, opened 2026-08-18
Keep the name from Create token on the grant

gvocale:
## Summary

Create token requires a name, then throws it away. The grant now stores that label and GetGrant returns it, so Settings can show the name you typed instead of only an ID.

## Why / context

The create form already collects a token name. Token exchange never forwarded it, and the grant record had no place to keep it. Caching the name in the page would vanish on refresh and would not help other browsers.

CLI and other unnamed grants stay unnamed. Empty name remains valid.

## What changed

- Token exchange accepts an optional name, the same kind of request-side extension as expires-in.
- GetGrant returns that name when one was stored.
- HTTP create forwards the typed name and echoes it in the create response instead of repeating the grant ID.
- Generated SDK clients carry the new fields. The Rust identity server also copies name onto the provider request and GetGrant response.
- The public-surface snapshot now includes Workflow ListRuns `definition_id` and `known_apps`. Those fields were already in the proto. Token name does not appear in that snapshot because Token posts the whole body.

## Validation

- [x] Automated: `go test ./internal/server -run 'TestCreateAPITokenForwardsExpiresIn|TestCreateAPITokenOmitsExpiresInWhenUnset'` from `gestaltd` (create response name and GetGrant name)
- [x] Automated: `go test ./internal/publicsurface -run TestManifestGolden` from `sdk/tools/sdkgen`
- [x] Automated: `cargo check --tests` from `sdk/rust`
- [x] Automated: `git diff --check`
- [ ] Manual: after this and the companion identity-provider plus Settings UI change are deployed, create a named token and confirm the list shows that name

## Reviewer focus

Please check that unnamed token exchange (CLI, session paths) still stores an empty name, and that GetGrant JSON includes `name` when present. The ListRuns golden refresh is proto that was already on main, not a Workflow behavior change in this PR.

## Rollout / compatibility

Merge this before the companion gestalt-providers PR. That PR pins `.github/gestalt-ref` to this branch so identity can compile against the new fields. After this squash-merges, retarget that pin to the merge commit.

Existing grants have no stored name. They will show as unnamed until people create new tokens after deploy.

## Evidence / demo

No rendered UI in this change. Proof is create forwarding the typed name, returning it on create, and GetGrant returning the same name from the grant-tracking stub.

## Links

Companion: https://github.com/valon-technologies/gestalt-providers/pull/1288


mzywang (approved):
Reviewed the token-name plumbing across the identity boundary and generated SDK surfaces. The create/GetGrant coverage and passing SDK checks look solid.

PR #3199, opened 2026-09-08
Fix large internal app gRPC responses

nika-gh:
## Summary

- Raise the internal Go app-provider client receive limit to 16 MiB.
- Apply the limit to local provider-process and hosted-runtime connection paths.
- Add regression coverage for the 4,438,910-byte response observed for the 3,312,204-byte Bradley workbook.

## Why

The Bradley RCM automation still failed in production after the provider-side attachment limit change because Gestalt rejected the response at the default gRPC receive limit:

`grpc: received message larger than max (4438910 vs. 4194304)`

This is a short-term transport unblock that gives headroom above the observed production response while keeping public gRPC defaults unchanged. Larger binary/app payloads should move to a streaming or handle-based operation instead of relying on larger unary gRPC responses.

## Validation

- Previous CI passed on the original transport change.
- Latest CI is pending after the 16 MiB reduction and comment cleanup.


mzywang (approved):
Looks good — raising the internal gRPC recv limit to 40 MiB for hosted-app clients (without changing public defaults) matches the Bradley Gmail attachment failure, and the large-response tests cover the production size.

nika-gh:
Closing this approach because Gestalt Core will keep the 4 MiB internal gRPC limit. We will pursue an app-level Bradley workaround that avoids returning the full attachment through one gRPC response.

mzywang (approved):
Re-approved after reducing the internal unary receive cap to 16 MiB and cleaning up the shared-code comment.

PR #3209, opened 2026-09-09
Reduce internal gRPC response limit to 16 MiB

mzywang:
## Summary
- reduce the short-term internal app-provider gRPC receive allowance from 40 MiB to 16 MiB
- update the comment to frame this as Bradley-production-response headroom rather than full Gmail 25 MiB coverage

## Validation
- not run; comment/constant-only change on top of #3199


mzywang:
Combined into #3199 directly.

PR #3243, opened 2026-09-13
[gestaltd] disconnect the requested credential before duplicate cleanup

mzywang:
## Description
Disconnect the explicitly requested credential before cleaning up duplicates for the same account. Previously, group ordering could select a different credential first and turn a best-effort duplicate deletion failure into a 500.

## Summary
- Prefer the requested instance within its matched account group.
- Make the existing regression test deterministic by giving the duplicate a newer timestamp.

## Test plan
- [x] Reproduced the previous failure in repeated local tests.
- [x] Disconnect and callback-origin tests passed 30 consecutive runs after the fix.


mzywang:

- #3243 ◀

PR #3244, opened 2026-09-14
[deploy] bound concurrent UI probes to fit startup admission

mzywang:
## Description
Evaluate the UI readiness inventory with at most eight concurrent probes so sequential request latency does not consume the instance startup window. Readiness still waits for every result and rejects the entire instance if any probe fails.

## Summary
- Bound probe concurrency and preserve deterministic report ordering.
- Publish a complete report only after all probes finish.

## Test plan
- [x] Readiness tests pass across five repetitions.
- [x] Regression test proves eight concurrent probes, no partial admission, ordered results, and rejection of a failed final probe.
- [x] Race-enabled readiness tests pass across three repetitions.


mzywang:

- #3244 ◀

